PR #81, opened 2024-08-19
Add Audit Log PVC

coreydaley:

- Fixes #_issue_number_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation
- [ ] If the bundle manifests have been updated, make sure to review the [`rhdh-operator.csv.yaml`](../.rhdh/bundle/manifests/rhdh-operator.csv.yaml) file accordingly

### DIFF
--- a/bundle/manifests/backstage-default-config_v1_configmap.yaml
+++ b/bundle/manifests/backstage-default-config_v1_configmap.yaml
@@ -15,6 +15,17 @@ data:
                   subject: legacy-default-config
                   # This is a default value, which you should change by providing your own app-config
                   secret: "pl4s3Ch4ng3M3"
+  audit-log-pvc.yaml: |
+    kind: PersistentVolumeClaim
+    apiVersion: v1
+    metadata:
+      name: audit-log-pvc # will be replaced
+    spec:
+      accessModes:
+        - ReadWriteOnce
+      resources:
+        requests:
+          storage: 1Gi
   db-secret.yaml: |-
     apiVersion: v1
     kind: Secret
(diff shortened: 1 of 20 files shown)


gazarenkov (changes requested):
Making a "per-case" (i.e. audit-log-pvc, dynamic-plugins-pvc etc) object in the model is not really what I would expect from the Operator.
Please take a look at https://issues.redhat.com/browse/RHIDP-3621 for more generic approach, where we can define and use PVC not making dedicated object per case.

config/rbac/role.yaml
```diff
@@ -24,6 +24,8 @@ rules:
   - persistentvolumeclaims
   - persistentvolumes
   verbs:
+  - create
+  - delete
```
durandom, line 28: @coreydaley does the operator delete any pvc/pv ? I guess it would be safer if we would not allow deletion of pvc, right?

bundle/manifests/backstage-default-config_v1_configmap.yaml
```diff
@@ -270,6 +284,8 @@ data:
               volumeMounts:
                 - mountPath: /opt/app-root/src/dynamic-plugins-root
                   name: dynamic-plugins-root
+                - mountPath: /var/log/audit
```
durandom, line 287: looking at https://github.com/janus-idp/backstage-showcase/blob/7880332d0e0919dcbc08a9dd040392e146d9f05f/packages/backend/src/logger/customLogger.ts#L70 the default path for the audit log 

Is the plan to guide the user via documentation to set the correct path for this volume to be used?
@schultzp2020 @Zaperex

coreydaley: I thought we had agreed upon /var/log/audit so not sure why it is different.

config/rbac/role.yaml
coreydaley: Testing this pull request did give me an error when it did not have the `delete` permission, so if we want to exclude it from ever being deleted that probably needs to be built into the code somewhere.

PR #83, opened 2024-08-20
chore: suspend renovate for go/k8s till 1.22 based UBI

gazarenkov:


rm3l (commented):
From https://github.com/redhat-developer/rhdh-operator/pull/85#issuecomment-2298187857, it looks like updates to `sigs.k8s.io/controller-runtime` also require Go 1.22.
Would it make sense to add a rule disabling dependency updates for `sig.k8s.io/**` as well?

PR #102, opened 2024-08-22
add dynamic-plugins-registry-auth secret

durandom:
For secure container registry we need a `auth.json` file that contains auth tokens. The file is stored as a secret like

```
oc create secret generic dynamic-plugins-registry-auth --from-file=~/.config/containers/auth.json
```

It is then picked up by `skopeo` during the dynamic plugin install process.

Accompanying PR for dynamic plugins:

https://github.com/janus-idp/backstage-showcase/pull/1479
https://github.com/redhat-developer/rhdh-chart/pull/41

### DIFF
--- a/config/manager/default-config/deployment.yaml
+++ b/config/manager/default-config/deployment.yaml
@@ -33,6 +33,11 @@ spec:
             defaultMode: 420
             optional: true
             secretName: dynamic-plugins-npmrc
+        - name: dynamic-plugins-registry-auth
+          secret:
+            defaultMode: 416
+            optional: true
+            secretName: dynamic-plugins-registry-auth
         - emptyDir: {}
           name: npmcacache
       initContainers:
@@ -61,6 +66,9 @@ spec:
               name: dynamic-plugins-npmrc
               readOnly: true
               subPath: .npmrc
+            - mountPath: /opt/app-root/src/.config/containers
+              name: dynamic-plugins-registry-auth
+              readOnly: true
             - mountPath: /opt/app-root/src/.npm/_cacache
               name: npmcacache
           workingDir: /opt/app-root/src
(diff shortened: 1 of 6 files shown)


config/manager/default-config/deployment.yaml
```diff
@@ -33,6 +33,11 @@ spec:
             defaultMode: 420
             optional: true
             secretName: dynamic-plugins-npmrc
+        - name: dynamic-plugins-registry-auth
+          secret:
+            defaultMode: 256
```
rm3l, line 38: Just curious about this. I just tested in my vanilla K8s cluster and, while the file permissions look good (per https://github.com/redhat-developer/rhdh-chart/pull/41#discussion_r1722797257), the problem is that the Secret is mounted as root by K8s. Since the container is running as non-root, the user is not able to read the file, which is owned by `root`. I guess we might run into this issue if the container process tries to access that file, no?

```
$ kubectl exec -it backstage-bs1-645f9b7fd-6tq9m -c install-dynamic-plugins -- /bin/sh 
sh-5.1$ ls -l /opt/app-root/src/.config/containers/auth.json
-r-------- 1 root root 147 Aug 22 15:18 /opt/app-root/src/.config/containers/auth.json
sh-5.1$ cat /opt/app-root/src/.config/containers/auth.json
cat: /opt/app-root/src/.config/containers/auth.json: Permission denied
sh-5.1$
```

durandom: good catch. i changed the defaultMode to `416` now, which results in`-rw-r----`. Tested it within my cluster and worked.

tests/e2e/testdata/backstage-operator-0.1.3.yaml
rm3l, line 1: This manifest file was generated using the `1.1.x` branch and is used to test the upgrade path from a previous version. I'd suggest reverting the changes to this file, since 1.1.x did not have support for this secret.

rm3l (commented):
Works better with the new value for `defaultMode` - thanks!
Just a small comment on the test resource below.

rm3l (approved):
Thanks for the changes.
Tested locally and in a cluster (both vanilla K8s and ClusterBot). Also tested upgrading from 1.2.x to this branch (with and without a dynamic plugins registry auth secret).

PR #113, opened 2024-08-27
Config docs (1.3) 

gazarenkov:
## Description

Configuration doc

https://issues.redhat.com/browse/RHIDP-1658

read and validate the doc


docs/configuration.md
```diff
@@ -1 +1,363 @@
...
+
+  ```
+  /my/path/app-config1.yaml
+  /my/path/app-config2.yaml
+  ```
+
+The ConfigMap key/value defines the file name and content, and these app-configs will be applied in the same order as they are declared in the CR. Therefore, as per this example, two ConfigMaps should be created in the namespace as a prerequisite. Then the Operator will create the files `/my/path/app-config1.yaml` and `/my/path/app-config2.yaml`, adding them to the end of the Backstage command-line arguments as follows:
+
+  ```
+  -app-config /my/path/app-config1.yaml -app-config /my/path/app-config2.yaml
```
rm3l, line 118: As depicted in https://github.com/redhat-developer/rhdh-operator/blob/main/api/v1alpha2/backstage_types.go#L92-L94, there is a limitation when using a single ConfigMap containing several filenames: the order in which they will translate into the container args is not guaranteed.
I think we should call this out explicitly here too. 
And also maybe update your example to use 2 different ConfigMaps, each with a single filename? (that's what you are doing by the way in the CR example where you are specifying 2 ConfigMaps in `spec.application.appConfig.configMaps`)

docs/configuration.md
```diff
@@ -1 +1,363 @@
...
+      subdomain: mysubdomain
+      tls:
+        certificate: "certificatecontent"
+        externalCertificateSecretName: "my-certificate"
+        key: "keycontent"
+        caCertificate: "caCertificatecontent"
+  ```
+  
+  Here, the user can specify some of the OpenShift Route specifications fields. The names of the Backstage `spec.application.route` fields correspond to the names of Route specifications fields and follow the same default rules if not specified.
+
```
rm3l, line 302: I'd suggest calling this out as well. See https://github.com/redhat-developer/rhdh-operator/blob/main/api/v1alpha2/backstage_types.go#L278-L279

```suggestion
  
  Also note that securing Routes with external certificates in TLS secrets (via the `spec.application.route.tls.externalCertificateSecretName` CR field) is a Technology Preview feature in OpenShift. It requires enabling the `RouteExternalCertificate` OpenShift Feature Gate and might not be functionally complete. See [Creating a route with externally managed certificate](https://docs.openshift.com/container-platform/4.16/networking/routes/secured-routes.html#nw-ingress-route-secret-load-external-cert_secured-routes) for more details.

```

rm3l (changes requested):
Thanks for these great docs. I added a few comments/suggestions.

docs/configuration.md
```diff
@@ -1 +1,363 @@
...
\ No newline at end of file
+# Configuration
+
+  It is highly recommended to read the [Design](design.md) document to understand the components that make up Backstage.
+
+The Backstage Operator supports three configuration levels: Default, Raw, and Custom Resource Config.
+
+## Default Configuration
+  
+  The Default Configuration defines the structure of all Backstage instances within a cluster. It consists of a set of YAML manifests that define Kubernetes resources for a Backstage instance. This configuration is located in the `*-default-configuration` ConfigMap in the Backstage operator namespace (usually called **backstage-system** or **backstage-operator**). For more details, refer to the [Admin Guide](admin.md).
```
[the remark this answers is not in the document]
gazarenkov: I put *-default-configuration on purpose, with multi profile the prefix can be different (depends on namePrefix), see for example rhdh profile, it called  rhdh-default-config b/c namePrefix: rhdh-.
So, it is for the "near" future (pretty sure we'll forget to modify it :) )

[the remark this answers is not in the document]
rm3l: Ok. Let's change the namespace name then ;)

```suggestion
  The Default Configuration defines the structure of all Backstage instances within a cluster. It consists of a set of YAML manifests that define Kubernetes resources for a Backstage instance. This configuration is located in the `*-default-configuration` ConfigMap in the Backstage operator namespace (usually called **backstage-system** or **rhdh-operator**). For more details, refer to the [Admin Guide](admin.md).
```

docs/configuration.md
gazarenkov: good point, fixing it

docs/configuration.md
```diff
@@ -1 +1,392 @@
...
+        - name: my-app-config
+  ```
+
+In this example, additional app-config is defined in the ConfigMaps called **my-app-config**, and it is mounted as YAML files to the `/my/path` directory in the Backstage container:
+
+  ```
+  /my/path/my-app-config.yaml
+  ```
+
+The ConfigMap key/value defines the file name and content, and these app-configs will be applied in the same order as they are declared in the CR. Therefore, as per this example, two ConfigMaps should be created in the namespace as a prerequisite. Then the Operator will create the files `/my/path/app-config1.yaml` and `/my/path/app-config2.yaml`, adding them to the end of the Backstage command-line arguments as follows:
```
rm3l, line 138: > Therefore, as per this example, two ConfigMaps should be created in the namespace as a prerequisite. Then the Operator will create the files `/my/path/app-config1.yaml` and `/my/path/app-config2.yaml`, adding them to the end of the Backstage command-line arguments as follows:

These sentences no longer seem consistent with the example provided. I guess they can be removed.

docs/configuration.md
rm3l: This was marked as resolved, but I couldn't see these lines. Does it make sense to call it out here?

rm3l:
Since this is for 1.3, and even if it is only upstream doc, I guess we should also cherry-pick this into the new [1.3.x branch](https://github.com/redhat-developer/rhdh-operator/tree/1.3.x).
cc @nickboldt @Fortune-Ndlovu

docs/configuration.md
gazarenkov: fixed

docs/configuration.md
gazarenkov: fixed

rm3l (approved):
Thanks for the changes!

/lgtm
/cherry-pick 1.3.x

PR #193, opened 2024-09-16
Image defined in the CR.spec should override the one defined in RELATED_IMAGE_backstage env variable

gazarenkov:
## Description
Image defined in the CR.spec should override the one defined in RELATED_IMAGE_backstage env variable

https://issues.redhat.com/browse/RHIDP-4013

## PR acceptance criteria

- [X] Tests

### DIFF
--- a/pkg/model/deployment_test.go
+++ b/pkg/model/deployment_test.go
@@ -16,6 +16,7 @@ package model
 
 import (
 	"context"
+	"os"
 	"testing"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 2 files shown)


nickboldt:
as a non-fluent go reader, I'm not sure what all the changes above do (I'd prefer more comments in the code) but if @rm3l likes this then +1 from me too.

/lgtm

pkg/model/deployment_test.go
```diff
@@ -169,6 +170,35 @@ spec:
 	assert.Equal(t, "my-vol", model.backstageDeployment.deployment.Spec.Template.Spec.Volumes[3].Name)
 }
 
+func TestImageInCRPrevailsOnEnvVar(t *testing.T) {
+	bs := *deploymentTestBackstage.DeepCopy()
+	bs.Spec.Deployment = &bsv1.BackstageDeployment{}
+	bs.Spec.Deployment.Patch = &apiextensionsv1.JSON{
```
rm3l, line 176: IMO, this test makes sense but I am missing a test that mimics the repro steps depicted in https://issues.redhat.com/browse/RHIDP-4013, i.e., using `bs.Spec.Application.Image` instead.
Here the conditions are different to me, and do not guarantee that the exact issue reported will not occur again.

Could we change this test (or add a separate test) using `bs.Spec.Application.Image` instead? And also make sure that both the init and main containers are set using this field (this is the current behavior expected from the `spec.application.image` field, which I recall was tested in the past before being removed at some point).

pkg/model/deployment_test.go
```diff
@@ -169,6 +170,35 @@ spec:
...
+spec:
+ template:
+   spec:
+     containers:
+       - name: backstage-backend
+         image: cr-image
+`),
+	}
+
+	os.Setenv(BackstageImageEnvVar, "envvar-image")
```
rm3l, line 187: Recommended to use `t.Setenv` instead to make sure this temporary env var value is unset automatically after the test, so it does not interfere with others trying to use the same environment variable.

gazarenkov: makes sense

pkg/model/deployment_test.go
gazarenkov: This test the issue we had I believe, the problem was not in bs.Spec.Application.Image, it was env var vs CR config.
So, I'd avoid to make extra (duplicating) tests for deprecated functionality

PR #281, opened 2024-10-01
chore: remove Licensing headers from individual files consistently for Operator

Fortune-Ndlovu:

https://issues.redhat.com/browse/RHIDP-4252?filter=-1

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation
- [ ] If the bundle manifests have been updated, make sure to review the [`rhdh-operator.clusterserviceversion.yaml`](../.rhdh/bundle/manifests/rhdh-operator.clusterserviceversion.yaml) file accordingly


Makefile
```diff
@@ -101,18 +101,10 @@ help: ## Display this help.
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
-.PHONY: generate
-generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
```
rm3l, line 105: Why has this target been removed? I think some other targets may depend on it.

Fortune-Ndlovu: I mean to only remove the `object:headerFile` option, not the entire target.

LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2023 Red Hat Inc.
+   Copyright 2024 Red Hat Inc.
```
rm3l, line 189:
```suggestion
   Copyright Red Hat
```

I just found an article titled `Copyright Notices for Open Source Projects` from the Legal team on The Source. It clearly does not make it mandatory to include the year:

> [...]
> We do not recommend inclusion of the year in the copyright notice. Attempting to consistently maintain dates in copyright notices is difficult, time-consuming, and is unnecessary.

So I think we can just simplify and omit the year as well.

It also recommends "Red Hat" over "Red Hat, Inc."

rm3l (changes requested):
There is also an occurrence of the license at the bottom of the `README.md` file, which needs to be updated as well, I think.

gazarenkov (approved):
LGTM with Armel's note about year

rm3l (approved):
Thanks @Fortune-Ndlovu !

/lgtm

PR #338, opened 2024-10-15
Remove usage of `kube-rbac-proxy` and switch to the built-in `WithAuthenticationAndAuthorization` filter instead

rm3l:
## Description

This removes all usage of the `kube-rbac-proxy` sidecar container, which was used to protect the `/metrics` endpoint exposed by the controller. This ensured that only authorized users and service accounts could access potentially sensitive metrics data.
Instead, it is now recommended to leverage the built-in [`WithAuthenticationAndAuthorization`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/metrics/filters#WithAuthenticationAndAuthorization) filter which provides a similar authn/authz protection of this endpoint but without the need for an extra container.
More context in https://github.com/kubernetes-sigs/controller-runtime/issues/2073 and https://book.kubebuilder.io/reference/metrics#how-the-metrics-endpoint-can-be-protected-

- Fixes https://issues.redhat.com/browse/RHIDP-4236
- Obsoletes https://issues.redhat.com/browse/RHIDP-2830
- Obsoletes https://issues.redhat.com/browse/RHIDP-2298

## PR acceptance criteria

- [x] Tests
- [ ] Documentation
- [x] If the bundle manifests have been updated, make sure to review the [`rhdh-operator.clusterserviceversion.yaml`](../.rhdh/bundle/manifests/rhdh-operator.clusterserviceversion.yaml) file accordingly

From this PR:

- The Operator Deployment should now have a single container if you deploy it with or without OLM
- You can run the E2E test case added in https://github.com/redhat-developer/rhdh-operator/pull/338/files#diff-7d1fc1e232b55677582b51e2623aae6d8c15b5f0960ac6100fc7d733cb795dac and make sure it passes:
```sh
# Download Ginkgo locally if needed
$ make ginkgo
$ IMG=quay.io/rhdh-community/operator:0.4.0-pr-338 \
  ./bin/ginkgo -v -p -focus 'operator metrics endpoint' tests/e2e

[...]
Ran 3 of 9 Specs in 48.442 seconds
SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 6 Skipped

Ginkgo ran 1 suite in 50.800777439s
Test Suite Passed
```


config/profile/backstage.io/manager.yaml
```diff
@@ -71,24 +71,32 @@ spec:
...
+        ports:
+          - name: health
+            containerPort: 8081
+          - name: metrics
+            containerPort: 8443
         livenessProbe:
           httpGet:
             path: /healthz
-            port: 8081
+            port: health
```
gazarenkov, line 93: Since this alias is used for both health and rediness I'd suggest to rename it to something like "probe" instead

config/profile/rhdh/manager.yaml
```diff
@@ -71,7 +71,10 @@ spec:
       - command:
         - /manager
         args:
-        - --leader-elect
+          - --health-probe-bind-address=:8081
```
gazarenkov, line 74: "probe-bind-address" ?

config/profile/backstage.io/manager.yaml
rm3l: IMO, probe is too generic as a term, since it might have different meaning (like a probe for scanning the network, or for collecting logs, or even for providing remote control...).
The liveness and readiness probes defined here both aim at providing health checks on that same port. So "health" as port name would be a more meaningful alias IMO. Or maybe "health-probe" if you prefer?
I think keeping "health" in the name helps better understand the purpose of this port.

config/profile/rhdh/manager.yaml
rm3l: Same arguments as in https://github.com/redhat-developer/rhdh-operator/pull/338#discussion_r1803777201 to keep "health" in the name.
Also, `health-probe-bind-address` seems consistent with what's there in the manager `Options` struct: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/manager.go#L226-L229

rm3l:
Rebased and forced-push to fix Git conflicts reported.

config/profile/backstage.io/manager.yaml
gazarenkov: To me "health" and "readiness" have different meaning (it may be health but not ready for ex) but both are "probes".
Another option would be "liveness"

rm3l: > To me "health" and "readiness" have different meaning
> [...]
> Another option would be "liveness"
>
Not sure I get your point. We are talking about "liveness" and "readiness" probes, right?
And both are targetting the same port alias: [liveness](https://github.com/rm3l/redhat-developer-hub-operator/blob/RHIDP-4236--remove-kube-rbac-proxy-and-use-WithAuthenticationAndAuthorization-instead/config/profile/backstage.io/manager.yaml#L90-L93) / [readiness](https://github.com/rm3l/redhat-developer-hub-operator/blob/RHIDP-4236--remove-kube-rbac-proxy-and-use-WithAuthenticationAndAuthorization-instead/config/profile/backstage.io/manager.yaml#L96-L99)
Naming the port alias "liveness" would make it look like it is specific to the liveness probe and make it different from the port used by the readiness probe..

PR #369, opened 2024-10-25
Make the CI builds install script able to detect and work against clusters with hosted control planes

rm3l:
## Description

This PR makes the `.rhdh/scripts/install-rhdh-catalog-source.sh` script detect whether the OCP cluster it is running against has a hosted control plane (HyperShift, ROSA, IBM Cloud, ...) and adapts its behavior accordingly.
In such a situation where the control plane is hosted, we cannot actually rely on the ImageContentSourcePolicy/ImageDigestMirrorSet, which will be created but won't be propagated to the cluster nodes.
As a workaround, this script rebuilds a new index by replacing all references to the internal registries with their equivalents on quay.io. It also does the same thing with the operator bundles, because the manifests in them might also contain references to internal registries. It pushes those new images into the internal cluster image registry.

#82 provided a similar attempt based on mirroring all images (airgap use case), but I think we still need to also rebuild the operator bundles because the manifests in them still contain references to internal registries. Also, I was surprised to see that updating the global secret on hosted control planes does not get propagated to the cluster nodes.

NOTE: We might be able to reuse that logic to make the airgap script work on such clusters as well. But this can be done in a follow-up step. [2]

[1] https://issues.redhat.com/browse/RHIDP-4415

- Relates to https://issues.redhat.com/browse/RHIDP-3205
- Relates to https://issues.redhat.com/browse/RHIDP-2919

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation
- [ ] If the bundle manifests have been updated, make sure to review the [`rhdh-operator.clusterserviceversion.yaml`](../.rhdh/bundle/manifests/rhdh-operator.clusterserviceversion.yaml) file accordingly

1. [Skip if you already have a cluster] Start an OCP cluster with a hosted control plane, e.g. with `rosa create 4.16 2h` on ClusterBot, or an OCP cluster on IBM Cloud
2. `oc login` against your OCP cluster
3. Run the install script: `.rhdh/scripts/install-rhdh-catalog-source.sh --[next|latest] --install-operator rhdh`.
The CatalogSource and operator should be installed and running, which was not the case previously.

Screenshot from 2024-10-25 11-18-22

image

4. Create a Backstage CR and make sure it runs

image

NOTES:
- The script should behave as previously on regular OpenShift clusters, i.e. by relying on `ImageContentSourcePolicy` / `ImageDigestMirrorSet` resources to mirror internal images.
- ~This will not work if routes are exposed with untrusted certificates (case of clusters created via `launch 4.x`). There is a TLS verification failure when trying to pull from the internal image registry, and I couldn't find a way to make the cluster accept/ignore this certificate (I guess because the control plane is hosted).~ (Fixed with [`38ba118` (#369)](https://github.com/redhat-developer/rhdh-operator/pull/369/commits/38ba118d19aa4a8b8963c77dfd58581218846eff))

/cc @nickboldt @Fortune-Ndlovu @Omar-AlJaljuli

### DIFF
--- a/.rhdh/docs/installing-ci-builds.adoc
+++ b/.rhdh/docs/installing-ci-builds.adoc
@@ -2,8 +2,12 @@
 
 *Prerequisites*
 
-* You are logged in as an administrator on the OpenShift web console.
-* You have configured the appropriate roles and permissions within your project to create an application. See the link:https://docs.openshift.com/container-platform/4.17/applications/index.html[Red Hat OpenShift documentation on Building applications] for more details.
+* `oc`. See link:https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/cli_tools/openshift-cli-oc#cli-installing-cli_cli-developer-commands[Installing the OpenShift CLI].
+* You are logged in as an administrator using `oc login`. See link:https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/cli_tools/openshift-cli-oc#cli-logging-in_cli-developer-commands[Logging in to the OpenShift CLI] or link:https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/cli_tools/openshift-cli-oc#cli-logging-in-web_cli-developer-commands[Logging in to the OpenShift CLI using a web browser].
+* `skopeo`. See link:https://github.com/containers/skopeo/blob/main/install.md[Installing Skopeo].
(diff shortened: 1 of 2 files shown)


rm3l:
@Omar-AlJaljuli FYI

Fortune-Ndlovu (approved):
/lgtm

Nice work on adapting the script to hosted control planes! the script looks solid, and I think this approach would be a useful adaptation for other airgapped setups.

PR #372, opened 2024-10-31
chore: remove ownruntime flag

gazarenkov:
Remove ownRuntime flag as it is not used

https://issues.redhat.com/browse/RHIDP-3609

## PR acceptance criteria

- [x] Tests

### DIFF
--- a/pkg/model/runtime.go
+++ b/pkg/model/runtime.go
@@ -88,7 +88,7 @@ func registerConfig(key string, factory ObjectFactory, multiple bool) {
 }
 
 // InitObjects performs a main loop for configuring and making the array of objects to reconcile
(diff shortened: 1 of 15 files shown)


pkg/model/runtime.go
rm3l, line 1: There is still an occurrence, even if it is just a comment. Can we remove that comment as well? Thanks.

```
pkg/model/runtime.go:146                   //setMetaInfo(backstageObject, backstage, ownsRuntime, scheme)
```

https://github.com/gazarenkov/rhdh-operator/blob/d70b34ce7927d1c0bd52c95ae9bc19cbca245b36/pkg/model/runtime.go#L146

rm3l (approved):

pkg/model/runtime.go
gazarenkov: removed

PR #379, opened 2024-11-05
Add workingDir support for mountPath and refactor mountPath applying

gazarenkov:
Add workingDir support for mountPath and refactor mountPath applying, add tests and docs

https://issues.redhat.com/browse/RHIDP-4784

## PR acceptance criteria

- [x] Tests
- [x] Documentation


api/v1alpha3/backstage_types.go
```diff
@@ -142,38 +141,38 @@ type ExtraFiles struct {
...
 	// List of references to Secrets objects mounted as extra files under the MountPath specified.
 	// For each item in this array, a key must be specified that will be mounted as a file.
 	// +optional
-	Secrets []FileObjectKeyRef `json:"secrets,omitempty"`
+	Secrets []FileObjectRef `json:"secrets,omitempty"`
 
 	// List of references to Persistent Volume Claim objects mounted as extra files
 	// For each item in this array, a key must be specified that will be mounted as a file.
 	// +optional
-	Pvcs []PvcRef `json:"pvcs,omitempty"`
```
rm3l, line 155: I understand that the refactoring, but this would expose the `key` field, which IMO does not make sense on PVCs. I find it more confusing for the CR users, especially when inspecting the documentation of the fields or when using the OCP console CR editor.
image

rm3l: Also, I think keeping the `PvcRef` struct would allow later to implement an OLM descriptor on the name field, so that users can pick PVCs from an auto-populated list of PVCs - scope of https://issues.redhat.com/browse/RHIDP-3113

rm3l (approved):

PR #393, opened 2024-11-14
fix auth/providers configuration issue

gazarenkov:
It fixes Backstage deployment crash with:

`/opt/app-root/src/node_modules/@backstage/backend-app-api/dist/wiring/BackendInitializer.cjs.js:243
              throw new errors.ForwardedError(
                    ^
ForwardedError: Plugin 'app' startup failed; caused by Error: Invalid app bundle schema. If this error is unexpected you need to run `yarn build` in the app. If that doesn't help you should make sure your config schema is correct and rebuild the app bundle again. Caused by the following schema error, Error: Config validation failed, Config must have required property 'providers' { missingProperty=providers } at /auth
    at /opt/app-root/src/node_modules/@backstage/backend-app-api/dist/wiring/BackendInitializer.cjs.js:243:21
`
on :next showcase image if auth/providers is not configured explicitly.


config/profile/rhdh/default-config/app-config.yaml
```diff
@@ -12,6 +12,10 @@ data:
               subject: legacy-default-config
               # This is a default value, which you should change by providing your own app-config
               secret: "pl4s3Ch4ng3M3"
+    auth:
+      providers:
+        github: {}
```
rm3l, line 17: I guess we can only have `auth.providers: {}` (without `github`) as the actual provider should not matter, according to the error message. It will also be consistent with the Helm Chart: https://github.com/redhat-developer/rhdh-chart/blob/main/charts/backstage/values.yaml#L53-L54

rm3l: I've gone ahead and pushed this change to unblock the failing nightly checks: https://github.com/redhat-developer/rhdh-operator/actions/workflows/nightly.yaml

rm3l (approved):
/lgtm

PR #582, opened 2024-12-19
Configure other than Backstage container to mount PVC volume to

gazarenkov:

https://issues.redhat.com/browse/RHIDP-4595

## PR acceptance criteria

- [X] Tests
- [X] Documentation
- [X] If the bundle manifests have been updated, make sure to review the [`rhdh-operator.clusterserviceversion.yaml`](../.rhdh/bundle/manifests/rhdh-operator.clusterserviceversion.yaml) file accordingly

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -42,7 +42,7 @@ Some objects, such as: app-config, configmap-files, secret-files, dynamic-plugin
 
 #### Object annotation for mounting a volume to a specific path
 
-Using **rhdh.redhat.com/mount-path** annotation it is possible to define the directory where **PersistentVolumeClaim** object will be mounted to Backstage Container.
+Using **rhdh.redhat.com/mount-path** annotation it is possible to define the directory where **PersistentVolumeClaim** object will be mounted.
 
 ```yaml
 apiVersion: v1
@@ -55,6 +55,28 @@ metadata:
 ```
   
 In the example above the PVC called **myclaim** will be mounted to **/mount/path/from/annotation** directory
+
+#### Object annotation for mounting a volume to specific container(s)
+
+Using **rhdh.redhat.com/containers** annotation it is possible to define the containers where **PersistentVolumeClaim** object will be mounted.
+
+Options:
+
+* No or empty annotation means the volume will be mounted to the Backstage container only
+* \* (asterik) means the volume will be mounted to all the containers
+* Otherwise comma separated names of container will be used
+
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: myclaim
+  annotations:
+    rhdh.redhat.com/mount-path: "init-dynamic-plugins,backstage-backend"
+...
+```
+In the example above the PVC called **myclaim** will be mounted to **init-dynamic-plugins** and **backstage-backend** containers
+
 
 ### Metadata Generation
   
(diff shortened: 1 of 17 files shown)


docs/configuration.md
```diff
@@ -56,6 +56,28 @@ metadata:
...
+* \* (asterik) means the volume will be mounted to all the containers
+* Otherwise comma separated names of container will be used
+
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: myclaim
+  annotations:
+    rhdh.redhat.com/mount-path: "init-dynamic-plugins,backstage-backend"
```
rm3l, line 75:
```suggestion
    rhdh.redhat.com/containers: "init-dynamic-plugins,backstage-backend"
```

docs/configuration.md
```diff
@@ -56,6 +56,28 @@ metadata:
...
 
+#### Object annotation for mounting a volume to specific container(s)
+
+Using **rhdh.redhat.com/containers** annotation it is possible to define the containers where **PersistentVolumeClaim** object will be mounted.
+
+Options:
+
+* No or empty annotation means the volume will be mounted to the Backstage container only
+* \* (asterik) means the volume will be mounted to all the containers
+* Otherwise comma separated names of container will be used
```
rm3l, line 67:
```suggestion
* Otherwise comma-separated names of containers will be used
```

docs/configuration.md
```diff
@@ -56,6 +56,28 @@ metadata:
   
 In the example above the PVC called **myclaim** will be mounted to **/mount/path/from/annotation** directory
 
+#### Object annotation for mounting a volume to specific container(s)
+
+Using **rhdh.redhat.com/containers** annotation it is possible to define the containers where **PersistentVolumeClaim** object will be mounted.
```
rm3l, line 61: This implies that the mount path will be the same in all the containers listed here, right?
What if I want to mount the PVC under specific paths for each container? IMO, we should handle this.
Thinking out loud, but how about extending the behavior of the `rhdh.redhat.com/mount-path` annotation above to include the container names, e.g.: `rhdh.redhat.com/mount-path: 'install-dynamic-plugins:/mount/path1,backstage-backend:/mount/path2'`?

docs/configuration.md
```diff
@@ -42,7 +42,7 @@ Some objects, such as: app-config, configmap-files, secret-files, dynamic-plugin
 
 #### Object annotation for mounting a volume to a specific path
 
-Using **rhdh.redhat.com/mount-path** annotation it is possible to define the directory where **PersistentVolumeClaim** object will be mounted to Backstage Container.
+Using **rhdh.redhat.com/mount-path** annotation it is possible to define the directory where **PersistentVolumeClaim** object will be mounted.
```
rm3l, line 45: The `pvcs.yaml` description in the table above should also be updated: https://github.com/redhat-developer/rhdh-operator/pull/582/files#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180L29

docs/configuration.md
gazarenkov: Thought about it (and discussed with Tomas)... Can not see the case when it should be really important for *default* configuration. We can (and will when needed) make it more flexible for CR but I'd really avoid scripting in annotations unless really unavoidable.

rm3l (approved):

PR #609, opened 2024-12-30
Default secrets

gazarenkov:
## Description
Added possibility for default secret to configure:

- several secrets (like PVCs)
- per-secret secret's mountPath 
- containers the secret should be mounted to (not only Backstage container)

https://issues.redhat.com/browse/RHIDP-4785

## PR acceptance criteria

- [x] Tests
- [x] Documentation


docs/configuration.md
```diff
@@ -56,6 +66,55 @@ metadata:
...
+kind: Secret
+metadata:
+  name: mysecret
+  annotations:
+    rhdh.redhat.com/mount-path: /mount/path/from/annotation
+...
+```
+In the example above the Secret called **mysecret** will be mounted to **/mount/path/from/annotation** directory
+
+#### rhdh.redhat.com/containers for mounting volume to specific container(s)
```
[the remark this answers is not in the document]
gazarenkov: Yes, do not merge it before 582. Put on hold label

pkg/model/deployment.go
```diff
@@ -52,13 +59,17 @@ func DeploymentName(backstageName string) string {
...
 	for i, c := range bsd.Spec.Template.Spec.Containers {
-		if c.Name == "backstage-backend" {
+		if c.Name == BackstageContainerName() {
 			return i
 		}
 	}
 	return -1
 }
 
+func BackstageContainerName() string {
```
rm3l, line 69: Since this returns the same string, could we use a simple constant instead (for better compiler optimization)?

gazarenkov: It used to be, I converted it in purpose.
The reason is that we might want to change the way we recognize backstage container name, for example put it as a deployment's lable or so...

rm3l (approved):

PR #664, opened 2025-01-16
Patch Operator default configuration with Platform specifics

gazarenkov:
## Description

This PR:

- Adds ability to provide optional per-profile/platform/object default configuration path as the additional file like: default-config/<object-name>.yaml.<platformId>. Where platformId is ocp or k8s (can be extended potentially)
- On start operator recognizes the platformId and applies corresponding patches
- RHDH profile provided with deployment.yaml.k8s, db-statefulset.yaml.k8s, service.yaml.k8s patches to make it possible to install RHDH on AKS, EKS, GKE w/o manual patching 
- Docs and tests added

https://issues.redhat.com/browse/RHIDP-1560
https://issues.redhat.com/browse/RHIDP-5587

## PR acceptance criteria

- [x] Tests
- [x] Documentation


config/profile/rhdh/default-config/db-statefulset.yaml.k8s
rm3l, line 1: Can this file be renamed to keep the `.yaml` file extension instead? It makes it easier to read/review with automatic color highlighting.

rm3l: Also, shouldn't this (along with the other files) be added to the `kustomization.yaml` [CM generator](https://github.com/redhat-developer/rhdh-operator/blob/main/config/profile/rhdh/kustomization.yaml#L34-L45) as well? Not tested, but I am not sure this would otherwise work if I deploy the operator bundle.

.rhdh/scripts/install-rhdh-catalog-source.sh
```diff
@@ -335,10 +335,10 @@ function update_refs_in_iib_bundles() {
...
             sed -i 's#registry-proxy.engineering.redhat.com/rh-osbs/rhdh-#quay.io/rhdh/#g' "$file"
 
-            if [[ "${IS_OPENSHIFT}" != "true" ]]; then
-              k8s_add_fsGroup_to_bundle_manifest "$file"
-              k8s_update_service_type_in_bundle_manifest "$file"
-            fi
+#            if [[ "${IS_OPENSHIFT}" != "true" ]]; then
+#              k8s_add_fsGroup_to_bundle_manifest "$file"
+#              k8s_update_service_type_in_bundle_manifest "$file"
+#            fi
```
rm3l, line 341: Let's remove this (along with the useless functions too) if not needed anymore.

config/profile/rhdh/default-config/db-statefulset.yaml.k8s
gazarenkov: > Can this file be renamed to keep the .yaml file extension instead?

I would like to keep it simple, adding "k8s" somewhere in between makes the logic of testing the name more sophisticated and I am not sure it worth so since it is most probably (at least as I can see it) will not be used for customized (customer's) configuration, only for default we distribute.
And it is colored if you say goland that this extension is yaml file 
Untitled

> be added to the kustomization.yaml
done

.rhdh/scripts/install-rhdh-catalog-source.sh
gazarenkov: OK

config/profile/rhdh/default-config/db-statefulset.yaml.k8s
rm3l: > And it is colored if you say goland that this extension is yaml file

Sure, but I was thinking of viewing the files from the GH web UI.
Anyway, I understand why it's done this way. Let's keep it simple, hoping that these patches won't need to grow too big.

rm3l (approved):
LGTM. Is there a follow-up JIRA to remove the manual steps from the product docs (GKE, EKS, AKS)?

PR #892, opened 2025-03-17
feat: Auto-detect the OCP ingress domain and populate the baseUrls in the default app-config ConfigMap [RHIDP-6192]

rm3l:
## Description

This PR automatically populates the default `app.baseUrl`, `backend.baseUrl` and `backend.cors.origin` fields on OCP, depending on the cluster ingress domain and the `spec.application.route.{enabled,host,subdomain}` fields.

See https://issues.redhat.com/browse/RHIDP-6192 for more context.

- Fixes https://issues.redhat.com/browse/RHIDP-6192

## PR acceptance criteria

- [x] Tests
- [x] Documentation

#### OCP

- Deploy the Operator using the image from this PR:
```
make deploy IMG=quay.io/rhdh-community/operator:0.6.0-pr-892-eccdd9a
```

- Create an empty CR without any customization:
```
oc apply -f examples/bs1.yaml
```

- Look at the default app-config CM and see that the values of the `app.baseUrl`, `backend.baseUrl` and `backend.cors.origin` have been updated, based on the cluster ingress domain:

<details>
<summary>oc describe cm backstage-appconfig-bs1</summary>

```yaml                                                                                                                                                                                       
app:
  # Default value. Might be updated on OCP based on the ingress domain and spec.application.route field
  baseUrl: https://backstage-bs1-my-ns.apps.rosa.vhyos-mx5yw-keo.1b9r.p3.openshiftapps.com
backend:
  auth:
    externalAccess:
      - type: legacy
        options:
          subject: legacy-default-config
          # This is a default value, which you should change by providing your own app-config
          secret: "pl4s3Ch4ng3M3"
  # Default value. Might be updated on OCP based on the ingress domain and spec.application.route field
  baseUrl: https://backstage-bs1-my-ns.apps.rosa.vhyos-mx5yw-keo.1b9r.p3.openshiftapps.com
  cors:
    # Default value. Might be updated on OCP based on the ingress domain and spec.application.route field
    origin: https://backstage-bs1-my-ns.apps.rosa.vhyos-mx5yw-keo.1b9r.p3.openshiftapps.com
auth:
  providers: {}

```

</details>

- If you open the app in the browser and log in as guest, you would notice that the home page is displayed out of the box. Previously, it would have been a 404 page.

<details>
<summary>Screenshot</summary>

image

</details>

#### K8s

- No change on vanilla K8s (still a 404 on the home page, until the user provides their own app-config in the CR):

<details>
<summary>Details</summary>

- Default app-config ConfigMap:
```yaml
app:
  # Default value. Might be updated on OCP based on the ingress domain and spec.application.route field
  baseUrl: http://localhost:7007
backend:
  auth:
    externalAccess:
      - type: legacy
        options:
          subject: legacy-default-config
          # This is a default value, which you should change by providing your own app-config
          secret: "pl4s3Ch4ng3M3"
  # Default value. Might be updated on OCP based on the ingress domain and spec.application.route field
  baseUrl: http://localhost:7007
  cors:
    # Default value. Might be updated on OCP based on the ingress domain and spec.application.route field
    origin: http://localhost:7007
auth:
  providers: {}
```

- Screenshot:

image

</details>


pkg/model/appconfig.go
```diff
@@ -78,11 +82,65 @@ func (b *AppConfig) addToModel(model *BackstageModel, _ bsv1.Backstage) (bool, e
...
+		}
+		if domain == "" {
+			klog.V(1).Info("no cluster ingress domain found, skipping setting the default baseUrls")
+			return nil
+		}
+		host += fmt.Sprintf(".%s", domain)
+	}
+	baseUrl := fmt.Sprintf("https://%s", host)
+	defaultAppConfig := b.ConfigMap.Data[defaultAppConfigFileName]
+	if defaultAppConfig == "" {
```
gazarenkov, line 136: It means that if user set the key other than "default.app-config.yaml" in the profile it won't work, right?

pkg/model/appconfig.go
```diff
@@ -78,11 +82,65 @@ func (b *AppConfig) addToModel(model *BackstageModel, _ bsv1.Backstage) (bool, e
...
+		return nil
+	}
+	b.ConfigMap.Data[defaultAppConfigFileName] = strings.ReplaceAll(defaultAppConfig, "http://localhost:7007", baseUrl)
```
gazarenkov, line 139: Do we really expect only http://localhost:7007 everywhere?
It is not hardcoded in a Backstage, should we?

pkg/model/appconfig.go
```diff
@@ -78,11 +82,65 @@ func (b *AppConfig) addToModel(model *BackstageModel, _ bsv1.Backstage) (bool, e
...
+	}
 	updatePodWithAppConfig(m.backstageDeployment, m.backstageDeployment.container(), b.ConfigMap.Name, m.backstageDeployment.defaultMountPath(), "", true, maps.Keys(b.ConfigMap.Data))
 	return nil
 }
 
+// updateDefaultBaseUrls detects if it is running on OCP and tries to set the baseUrl in the default app-config,
+// per the cluster ingress domain and the Route spec.
+// Note that this is purposely done on a best effort basis. So it is not considered an issue if the cluster ingress domain
+// could not be determined, since the user can always set it explicitly in their custom app-config.
+func (b *AppConfig) updateDefaultBaseUrls(backstage bsv1.Backstage) error {
```
gazarenkov, line 98: I think this function has a lot of responsibilities.
Let's consider some separation of concerns by splitting it leaving app-config update part here and moving recognizing of route url to the route object. 
This way it is not needed to duplicate validations like isRouteEnabled and isOpenshift here because it is a prerogative of Route (which is Openshift object by definition).

pkg/utils/utils.go
```diff
@@ -295,3 +300,39 @@ func getPlatform() string {
 	}
 	return PlatformK8s
 }
+
+// GetOCPIngressDomain returns the OpenShift Ingress domain
+func GetOCPIngressDomain() (string, error) {
```
gazarenkov, line 305: This call to be moved level up, to controller.
The simplest option I can see is to extend ExternalConfig with ingress domain field  and make it filled by spec-preprocessor (same as other ExternalConfig fields)
This way it is possible to add/move the testing of logic to unit tests instead of integration tests.

docs/configuration.md
```diff
@@ -518,10 +518,25 @@ spec:
         caCertificate: "caCertificatecontent"
 ```
   
-Here, the user can specify some of the OpenShift Route specifications fields. The names of the Backstage `spec.application.route` fields correspond to the names of Route specifications fields and follow the same default rules if not specified.
+Here, the user can specify some of the OpenShift Route specifications fields. The names of the Backstage `spec.application.route` fields correspond to the names of Route specification fields and follow the same default rules if not specified.
 
 Also note that securing Routes with external certificates in TLS secrets (via the `spec.application.route.tls.externalCertificateSecretName` CR field) is a Technology Preview feature in OpenShift. It requires enabling the `RouteExternalCertificate` OpenShift Feature Gate and might not be functionally complete. See [Creating a route with externally managed certificate](https://docs.openshift.com/container-platform/4.16/networking/routes/secured-routes.html#nw-ingress-route-secret-load-external-cert_secured-routes) for more details.
-  
+
+##### Default base URLs
```
gazarenkov, line 525: It has to be a part of "Default configuration" section I believe?

docs/configuration.md
```diff
@@ -518,10 +518,25 @@ spec:
...
+- No change if `spec.application.route.enabled` is explicitly set to `false`
+- The base URLs are set to `https://<spec.application.route.host>` if `spec.application.route.host` is set in the Backstage CR
+- The base URLs are set to `https://<spec.application.route.subdomain>.<cluster_ingress_domain>` if `spec.application.route.subdomain` is set in the Backstage CR. `<cluster_ingress_domain>` is the OpenShift ingress domain. See [The Ingress configuration asset](https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/networking/networking-operators#nw-ne-openshift-ingress_configuring-ingress) for more details. 
+- The base URLs are set to `https://backstage-<CR_name>-<namespace>.<cluster_ingress_domain>`, which is the domain set by default for the Route created by the Operator. `<cluster_ingress_domain>` is the OpenShift ingress domain. See [The Ingress configuration asset](https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/networking/networking-operators#nw-ne-openshift-ingress_configuring-ingress) for more details.
+
+Note that this behavior is done on a best-effort basis and only on OpenShift.
```
gazarenkov, line 536: We probably have to be more clear about limitations/expectations here (like expected name of app-config key and hardcoded default localhost:7007)?

gazarenkov (changes requested):
Nice feature.

IMO the issue is that we hardcode the configurable things (like existence default.app-config.yaml key) which is ok for demo but may lead to potential issues in practice.
As minimim we have to clearly document the limitations but ideally I would prefer to have a proper model of app-config (which I see as a part of https://issues.redhat.com/browse/RHIDP-2269)
Other comments inline.

pkg/model/appconfig.go
rm3l: Good point. I'll update this.

pkg/model/appconfig.go
rm3l: This is the default value in both [upstream Backstage image](https://github.com/backstage/backstage/blob/master/packages/create-app/templates/default-app/app-config.production.yaml) and [RHDH](https://github.com/redhat-developer/rhdh/blob/main/app-config.example.production.yaml).
That's why I thought it would work when deploying Backstage/RHDH images.
But I can try to make it generic enough to update all those fields regardless.

rm3l:
> IMO the issue is that we hardcode the configurable things (like existence default.app-config.yaml key) which is ok for demo but may lead to potential issues in practice. As minimim we have to clearly document the limitations.

@gazarenkov Thanks for your feedback on this. I've removed the hardcoded assumptions made initially (file name in the default app-config ConfigMap, default base URL), and updated the docs and tests accordingly. PTAL.


> but ideally I would prefer to have a proper model of app-config (which I see as a part of [issues.redhat.com/browse/RHIDP-2269](https://issues.redhat.com/browse/RHIDP-2269)) 

Sure. But based on the issues/discussions that were reported related to the baseUrls, I thought this quick PR would make things easier at least for OCP users, without having to touch the CRD.
Sure, I think [RHIDP-2269](https://issues.redhat.com/browse/RHIDP-2269) could be tackled subsequently and should not conflict with this. Good candidate for 1.7? ;-)

pkg/utils/utils.go
```diff
@@ -295,3 +300,35 @@ func getPlatform() string {
 	}
 	return PlatformK8s
 }
+
+// GetOCPIngressDomain returns the OpenShift Ingress domain
+func GetOCPIngressDomain(cl client.Client) (string, error) {
```
gazarenkov, line 305: Can we move it to controller level (it is called from there I think) to not to make REST calls from utils?

pkg/model/appconfig.go
```diff
@@ -78,17 +80,72 @@ func (b *AppConfig) addToModel(model *BackstageModel, _ bsv1.Backstage) (bool, e
 }
 
 // implementation of RuntimeObject interface
-func (b *AppConfig) updateAndValidate(m *BackstageModel, _ bsv1.Backstage) error {
+func (b *AppConfig) updateAndValidate(m *BackstageModel, backstage bsv1.Backstage) error {
+	b.updateDefaultBaseUrls(m, backstage)
```
gazarenkov, line 84: I think a better way is to contribute baseUrls by calling it from route.updateAndValidate method, pretty much like files/envvars objects contribute mounts and envvars to deployment.
Have you tried this way?

pkg/utils/utils.go
rm3l: Updated in [`d127f5f` (#892)](https://github.com/redhat-developer/rhdh-operator/pull/892/commits/d127f5f315288f14807d47545000dabd0f6b19cf)

pkg/model/appconfig.go
rm3l: PTAL at [`ccd6ff3` (#892)](https://github.com/redhat-developer/rhdh-operator/pull/892/commits/ccd6ff3d2b0d197e057f2bef84ce317e3c29e94d) and [`9765a32` (#892)](https://github.com/redhat-developer/rhdh-operator/pull/892/commits/9765a3219fc287f07a13e34ea1994a6f8b24686a)

rm3l:
/cherry-pick release-1.6

[RHIDP-6192](https://issues.redhat.com/browse/RHIDP-6192) is for 1.6.0

pkg/model/route.go
```diff
@@ -126,12 +131,97 @@ func (b *BackstageRoute) addToModel(model *BackstageModel, backstage bsv1.Backst
...
+	}
+}
+
+// buildBaseUrl returns the base URL that should be considered as default on OpenShift,
+// per the cluster ingress domain and the Route spec.
+func buildBaseUrl(model *BackstageModel, backstage bsv1.Backstage) string {
+	host := fmt.Sprintf("%s-%s", RouteName(backstage.Name), backstage.Namespace)
+	appendIngressDomain := true
+	if backstage.Spec.Application != nil && backstage.Spec.Application.Route != nil {
+		// Per the Route spec, if a user specifies both the host and subdomain, the host takes precedence.
```
gazarenkov, line 211: Would it make sense to put this info to doc?

gazarenkov (approved):
lgtm

pkg/model/route.go
rm3l: Not sure... We are already mentioning in the [Route configuration docs](https://github.com/redhat-developer/rhdh-operator/blob/main/docs/configuration.md#route) that the Route fields allow specifying `some of the OpenShift Route specifications fields.`
So I think people can refer to the Route spec for further details.

PR #991, opened 2025-04-04
chore: Fix the go-install-tool function in the Makefile

rm3l:
## Description
Sometimes, it would randomly fail to move the go-install'ed binary for some reason (probably the sed operation).

This is an attempt at solving this and making it more reliable.

&mdash;

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -503,8 +503,15 @@ define go-install-tool
 set -e; \
 package=$(2)@$(3) ;\
 echo "Downloading $${package}" ;\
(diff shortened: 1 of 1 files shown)


rm3l:
Actually, based on the output of https://github.com/redhat-developer/rhdh-operator/actions/runs/14262277350/job/39976333114?pr=991, the error is not in the `go-install-tool` function per se.

/close

PR #1001, opened 2025-04-07
feat: simplify rhdh-operator dockerfile

flacatus:

- Fixes #_issue_number_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation


flacatus:
cc @nickboldt

rm3l (commented):
Thanks for this PR, @flacatus !
Could you please rebase your PR branch to fix the PR build errors reported [here](https://github.com/redhat-developer/rhdh-operator/actions/runs/14306075044/job/40098823242?pr=1001)?
Also, I think [`.rhdh/docker/Dockerfile`](https://github.com/flacatus/rhdh-operator/blob/new_dockerfile/.rhdh/docker/Dockerfile) should be updated accordingly as well.

Dockerfile
```diff
@@ -16,20 +16,12 @@ RUN dnf -q -y update
...
 ENV CONTAINER_SOURCE=/opt/app-root/src
-WORKDIR /workspace
+WORKDIR $CONTAINER_SOURCE
 
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-COPY $EXTERNAL_SOURCE/go.mod ./go.mod
-COPY $EXTERNAL_SOURCE/go.sum ./go.sum
+# Copy the entire source code into the container first then download Go module dependencies.
+COPY $EXTERNAL_SOURCE $CONTAINER_SOURCE
```
rm3l, line 22: Could you expand a little bit on the motivations behind this PR? I feel like separating the `go.mod` and `go.sum` copy from the rest of the source code helps make the most of layer caching. Otherwise, the `go mod download` layer will be invalidated whenever there is a code change, even if the dependencies are not modified.

flacatus: @rm3l We would like to integrate Sealights in the operator. When we perform Sealights scanning in Konflux, Sealights generates an package that needs to be added to the build container runtime. That is why I put COPY . . to container. Do you have a better approach on go mod download?

rm3l (approved):
Thanks for the discussion around this. LGTM with the recent changes that preserve the `go mod download` cache optimization.

/lgtm

PR #1169, opened 2025-05-16
ci: Configure SeaLights to use Ginkgo for E2E tests

rm3l:
## Description
Our E2E tests leverage the Ginkgo framework.
Per [1], we should run the scan command with the `--tests-runner --enable-ginkgo` flags.

[1] https://sealights.atlassian.net/wiki/spaces/SUP/pages/4113891329/Using+Go+Agent+-+Running+your+Tests#Using-Ginkgo

Related to https://issues.redhat.com/browse/RHIDP-7360

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

/cc @npotluri-rh

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -106,8 +106,11 @@ jobs:
       - name: Configuring SeaLights
         if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
         run: |
-          echo "[SeaLights] Configuring SeaLights to scan for main branch nightly"
-          ./slcli config create-bsid --app rhdh-operator --branch ${{ matrix.branch }} --build "$BUILD_TIME:$LATEST_COMMIT_SHA"
+          echo "[SeaLights] Configuring SeaLights to scan for ${{ matrix.branch }} branch nightly"
+          ./slcli config create-bsid \
+            --app rhdh-operator \
+            --branch ${{ matrix.branch }} \
+            --build "$BUILD_TIME:$LATEST_COMMIT_SHA"
         env:
           LATEST_COMMIT_SHA: ${{ github.sha }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yaml
```diff
@@ -106,17 +106,30 @@ jobs:
...
+            --enable-ginkgo \
+            --bsid buildSessionId.txt \
+            --path-to-scanner ./slgoagent \
+            --workspacepath "./" \
+            --scm git \
+            --scmBaseUrl https://github.com/redhat-developer/rhdh-operator \
+            --scmProvider github
         env:
+          # Required if using Ginkgo: https://sealights.atlassian.net/wiki/spaces/SUP/pages/4113891329/Using+Go+Agent+-+Running+your+Tests#Using-Ginkgo
+          SEALIGHTS_IGNORE_GO_TESTS: "true"
```
gazarenkov, line 132: I am not sure skipping Go tests it is what we want (they cover Absolute majority of code)

Also,
"**If your tests are written exclusively using Ginkgo's** Domain-Specific Language (DSL) — which employs specialized syntax and constructs such as Describe, Context, and It — **and do not rely on the standard Go testing.T type, you can disable instrumentation**, reporting, and optimization of the testing.T type by the Sealights agent by setting the environment variable SEALIGHTS_IGNORE_GO_TESTS to true (default is false). The Ginkgo DSL Tests will remain instrumented." 
does not look like a requirement to disable testing.T type to enable Ginkgo scanning
I am not sure I understand what kind of analysis we want to get if we disable scanning of tests which cover about 60% (for the time being) functionality :)

.github/workflows/nightly.yaml
```diff
@@ -106,17 +106,30 @@ jobs:
       - name: Configuring SeaLights
         if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
         run: |
-          echo "[SeaLights] Configuring SeaLights to scan for main branch nightly"
-          ./slcli config create-bsid --app rhdh-operator --branch ${{ matrix.branch }} --build "$BUILD_TIME:$LATEST_COMMIT_SHA"
+          echo "[SeaLights] Configuring SeaLights to scan for ${{ matrix.branch }} branch nightly"
```
gazarenkov, line 109: Do we really need every night scanning for released branches?

gazarenkov (changes requested):

.github/workflows/nightly.yaml
rm3l: My understanding is that this instruments only the E2E tests, which are... written exclusively using Ginkgo's DSL. I agree our other unit and integration tests are using Go tests, but they are not run as part of these nightly E2E checks. So this still makes sense on the other PR integration SL with PRs.

Actually, I was a bit surprised by the overall coverage reported in the SL dashboard for these E2E tests: 8% (unless I misread it):

image

I want to see if these settings change anything in the SL dashboard. If it doesn't, I'm fine with reverting this.

.github/workflows/nightly.yaml
rm3l: IMO, yes, as long as they are supported.

.github/workflows/nightly.yaml
rm3l: Okay, according to @npotluri-rh , this was already tried and SL unfortunately didn't recognize those as tests, without any error.
Closing for now.

/close

PR #1177, opened 2025-05-17
chore: small testing related fixes

gazarenkov:
## Description
- add tidy step to pr.yaml to not failed on lint step due to inconsistent mod (usually in renovate PRs)
- exclude alpha1 and alpha2 packages from test covering
- add pne more test for plugin-deps

- [x] Tests


.github/workflows/pr.yaml
```diff
@@ -49,8 +48,12 @@ jobs:
         with:
           go-version-file: 'go.mod'
 
+      - name: Tidy
+        # run this step to ensure that the go.mod file is up to date, make lint will fail if it is not
+        run: go mod tidy
```
rm3l, line 53: But it is great that `make lint` fails if the `go.mod`/`go.sum` files are not up-to-date, no? This, IMO, is a reminder to run it locally and push the changes.
IMO, we should not run `go mod tidy` as a mutation step here, but we can definitely run to enforce cleanliness. Or if we want to run it in CI and take the resulting changes into account, we should maybe push the potential changes back to the branch.
As for Renovate PRs, there is already a post-update step that runs `go mod tidy` (added in https://github.com/redhat-developer/rhdh-operator/pull/543): https://github.com/redhat-developer/rhdh-operator/blob/main/.github/renovate.json#L23-L25

gazarenkov: OK, I was confused with failed jobs related to k8s version.
Do you think we can configure it to not to offer such PRs at all b/c it is really confusing?

I reverted, but, in general, I do not think the right way is that obvious, as I think we have to do it for prod anyway.

rm3l (approved):
/lgtm

PR #1191, opened 2025-05-22
chore(log): Warn if metrics are served over plainttext HTTP

Fortune-Ndlovu:
## Description
This adds a startup log warning when --metrics-secure=false is used,
to make it clear that exposing metrics without TLS/authn is only
intended for local development. Production bundles override this to true.

- Fixes https://issues.redhat.com/browse/RHIDP-7627

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -63,6 +63,10 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	if !secureMetrics {
+		setupLog.Info("Metrics are served over plaintext HTTP. This is only intended for local development.")
+	}
+
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	// if the enable-http2 flag is false (the default), http/2 should be disabled


cmd/main.go
```diff
@@ -63,6 +63,10 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	if !secureMetrics {
+		setupLog.Info("Metrics are served over plaintext HTTP. This is only intended for local development.")
```
rm3l, line 67: Note that the metrics service can be disabled altogether. In this case, it doesn't make sense to display this warning message.

rm3l (changes requested):

cmd/main.go
rm3l: Wondering if this would work if the logger has not been set. Did you check if this worked as intended? I suspect this block might need to be moved after the call to `ctrl.SetLogger` in L70.

Fortune-Ndlovu: Thanks for the heads-up! 
Also, small clarification: should we use `setupLog.Warn(...)` instead of `setupLog.Info(...)`  for this case? It seems like a better fit since it’s cautionary, even though it’s not an error. wdyt?

Fortune-Ndlovu: Oh yes, I see: This package restricts the logging API to just 2 types of logs: info and error.
Info logs are things you want to tell the user, which are not errors. Error logs are, well, errors

PR #1192, opened 2025-05-22
incorporate unit and integration tests into nightly

npotluri-rh:
## Description
Incorporated Unit and Integration test runs also into nightly

- Fixes #_RHIDP-7360_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

This is followup after first iteration of sealights integration to nightly workflow.
Trigger manual run to test after PR merged and observe the results on SeaLights Dashboard

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -7,7 +7,7 @@ on:
     - cron: "34 23 * * *"
 
 jobs:
-  e2e-tests:
+  nightly-tests:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -16,7 +16,7 @@ jobs:
           - main
           - release-1.6
           - release-1.5
-    name: 'E2E Tests on ${{ matrix.branch }}'
+    name: 'Nightly Tests on ${{ matrix.branch }}'
     concurrency:
       group: '${{ github.workflow }}-${{ matrix.branch }}'
       cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yaml
```diff
@@ -111,13 +111,53 @@ jobs:
...
         run: |
-          echo "[SeaLights] Running the SeaLights e2e tests scan"
+          echo "[SeaLights] Running the SeaLights scan"
           ./slcli scan --bsid buildSessionId.txt  --path-to-scanner ./slgoagent  --workspacepath "./" --scm git --scmBaseUrl https://github.com/redhat-developer/rhdh-operator --scmProvider github
-        env:
-          SEALIGHTS_TEST_STAGE: "E2E Tests Nightly"
+
+      # gosec needs a "build" stage so connect it to the lint step which we always do
+      - name: Build
+        if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
```
rm3l, line 122:
```suggestion
```

Since this does not rely on the operator image, I guess we can always run this step.

.github/workflows/nightly.yaml
```diff
@@ -111,13 +111,53 @@ jobs:
...
+        run: make lint
+
+      - name: Run Controller
+        if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
```
rm3l, line 126:
```suggestion
```

Same point, but it would require running the "Start Minikube" step as well, which should be fine.

.github/workflows/nightly.yaml
```diff
@@ -111,13 +111,53 @@ jobs:
...
+        # run this stage only if there are changes that match the includes and not the excludes
+        run: |
+          # Need to 'make install' first, so that the necessary tool binaries (like controller-gen) can be downloaded locally.
+          # Otherwise, we might end up with a race condition where the tool binary is not yet downloaded,
+          # but the `make test` command tries to use it.
+          make manifests generate fmt vet install
+          make run &
+
+      - name: Create a Unit Tests session
+        if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
```
rm3l, line 136:
```suggestion
```

Same point.

.github/workflows/nightly.yaml
```diff
@@ -111,13 +111,53 @@ jobs:
...
+        run: ./slcli test start-stage --bsid=buildSessionId.txt --testStage "Unit Tests"
+
+      - name: Test
+        if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
```
rm3l, line 140:
```suggestion
```

.github/workflows/nightly.yaml
```diff
@@ -111,13 +111,53 @@ jobs:
...
+        run: make test
+
+      - name: Create a Integration Tests session
+        if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
+        run: |
+          ./slcli test end-stage --bsid=buildSessionId.txt --executionId "Unit Tests"
+          ./slcli test start-stage --bsid=buildSessionId.txt --testStage "Integration Tests"
+
+      - name: Generic Integration test
+        if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
```
rm3l, line 151:
```suggestion
```

.github/workflows/nightly.yaml
```diff
@@ -111,13 +111,53 @@ jobs:
...
+        run: ./slcli test start-stage --bsid=buildSessionId.txt --testStage "Unit Tests"
+
+      - name: Test
+        if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
+        # run this stage only if there are changes that match the includes and not the excludes
+        run: make test
+
+      - name: Create a Integration Tests session
+        if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
```
rm3l, line 145:
```suggestion
```

.github/workflows/nightly.yaml
```diff
@@ -111,13 +111,53 @@ jobs:
...
+        run: |
+          ./slcli test end-stage --bsid=buildSessionId.txt --executionId "Unit Tests"
```
rm3l, line 147: Just curious, what happens if the `make test` step above fails and this "Create a Integration Tests session" step is not executed, meaning that `./slcli test end-stage` is never called?

npotluri-rh: Never called also ends eventually after timeout.
Explicitly calling end to start a new stage. we can skip it too as that is the default behavior.

.github/workflows/nightly.yaml
npotluri-rh: > Since this does not rely on the operator image, I guess we can always run this step.

So added condition on 'operator image' because E2E tests need it. I was thinking to run all three(unit,integration and e2e) or none as part of nightly. Before this there were no unit and Integration tests in nightly.

rm3l: Alright, makes sense.

.github/workflows/nightly.yaml
rm3l: Fine, let's see how it goes..

rm3l:
Triggered a new nightly workflow run here: https://github.com/redhat-developer/rhdh-operator/actions/runs/15192058484

PR #1284, opened 2025-06-19
fix: remove default replicas value to support HPA scaling

Fortune-Ndlovu:
## Description
This PR emoves the `replicas: 1` field from `config/profile/rhdh/default-config/deployment.yaml`.

This change allows users to apply Horizontal Pod Autoscalers (HPA) to the Backstage deployment without the operator reverting replica counts.

- Verified locally on Kind with metrics-server
- Confirmed HPA scales to 3 under CPU load
- Operator no longer resets to 1

You can view my entire replication of the issue and overall steps I took to ensure the fix works, in the issue comments section: https://issues.redhat.com/browse/RHIDP-7857

- Fixes https://issues.redhat.com/browse/RHIDP-7857 

## PR acceptance criteria

- [x] Tests
- [x] Documentation

### DIFF
--- a/config/profile/rhdh/default-config/db-statefulset.yaml
+++ b/config/profile/rhdh/default-config/db-statefulset.yaml
@@ -4,7 +4,7 @@ metadata:
   name: backstage-psql-cr1 # placeholder for 'backstage-psql-<cr-name>'
 spec:
   podManagementPolicy: OrderedReady
-  replicas: 1
+  # replicas: 1 # Intentionally omitted to allow HPA or custom scaling control.
   selector:
     matchLabels:
       rhdh.redhat.com/app: backstage-psql-cr1 # placeholder for 'backstage-psql-<cr-name>'
--- a/config/profile/rhdh/default-config/deployment.yaml
+++ b/config/profile/rhdh/default-config/deployment.yaml
@@ -3,6 +3,7 @@ kind: Deployment
 metadata:
   name: backstage # placeholder for 'backstage-<cr-name>'
 spec:
+  # replicas: 1 # Intentionally omitted to allow HPA or custom scaling control.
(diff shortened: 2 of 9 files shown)


config/profile/rhdh/default-config/deployment.yaml
```diff
@@ -3,7 +3,6 @@ kind: Deployment
 metadata:
   name: backstage # placeholder for 'backstage-<cr-name>'
 spec:
-  replicas: 1
```
rm3l, line 6: Could you comment this line instead and add an additional comment explaining why this is intentionally commented out? I think this will be helpful for our future selves.

rm3l (changes requested):

config/profile/rhdh/default-config/deployment.yaml
Fortune-Ndlovu: Sounds good!

gazarenkov (approved):

config/profile/rhdh/default-config/db-statefulset.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: backstage-psql-cr1 # placeholder for 'backstage-psql-<cr-name>'
 spec:
   podManagementPolicy: OrderedReady
-  replicas: 1
+  # replicas: 1 # Intentionally omitted to allow HPA or custom scaling control.
```
rm3l, line 7: [config/profile/backstage.io/default-config/db-statefulset.yaml](https://github.com/redhat-developer/rhdh-operator/blob/main/config/profile/backstage.io/default-config/db-statefulset.yaml) should also be updated.

rm3l (approved):
/lgtm

PR #1297, opened 2025-06-23
Allow enabling Orchestrator plugin on rhdh 

gazarenkov:
## Description

NOTE: it depends on https://github.com/redhat-developer/rhdh/pull/3016 (will be in draft until merge 3016)

https://issues.redhat.com/browse/RHIDP-7919

## PR acceptance criteria

- [x] Tests
- [x] Documentation

To test:
- make plugin-infra install run
- oc create namespace backstage
- oc apply -f examples/orchestrator.yaml -n backstage
Then: 
- make sure all 4 related pods are started on backstage namespace, 
- get url from the route and go there
- make sure Orchestrator link is workable

## Summary by Sourcery

Enable Orchestrator plugin on Red Hat Developer Hub by switching to Red Hat npm package references, refining dynamic plugin configuration, and updating related manifests, documentation, and tests.

Enhancements:
- Switch dynamic plugin package references to @redhat/backstage-plugin-orchestrator npm coordinates for frontend and backend modules.
- Remove deprecated dynamic-plugins-npmrc and registry-auth secret volume mounts across deployment and bundle configurations.
- Update NPM_CONFIG_USERCONFIG environment variable to reference the nested .npmrc file under .npmrc.dynamic-plugins.
- Add a Secret manifest for dynamic-plugins-npmrc in the default profile configuration.

Documentation:
- Clarify enabling Orchestrator plugin via npm.registry.redhat.com and update example configuration.

Tests:
- Update TestDefaultMultiSecretEnv to assert that generated secret names use the expected 'backstage-envs-<name>' prefix.

Chores:
- Bump createdAt timestamps in CSV manifests.
- Remove unused IsOpenShift field in operator controller setup.


config/profile/rhdh/default-config/deployment.yaml
```diff
@@ -28,16 +28,6 @@ spec:
...
-        - name: dynamic-plugins-npmrc
-          secret:
-            defaultMode: 420
-            optional: true
-            secretName: dynamic-plugins-npmrc
-        - name: dynamic-plugins-registry-auth
-          secret:
-            defaultMode: 416
-            optional: true
-            secretName: dynamic-plugins-registry-auth
```
rm3l, line 40: I guess removing this will break the ability to load plugins from private OCI registries, documented here: https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.6/html/installing_and_viewing_plugins_in_red_hat_developer_hub/assembly-third-party-plugins#proc-load-plugin-oci-image_assembly-install-third-party-plugins-rhdh

> For an Operator-based deployment, replace `<secret_name>` with `dynamic-plugins-registry-auth`.

gazarenkov: It does not break the ability
The way it is documented/implemented considers single dynamic-plugins-registry-auth secret per namespace, which (according to our concept) is not correct. 
It can be any secret (generated from default-config or user created and referenced in CR) of auth.json format mounted to ~/.config}/containers/auth.json

rm3l: > single dynamic-plugins-registry-auth secret per namespace, which (according to our concept) is not correct.

I thought we were recommending against multiple RHDH instances in a single namespace anyway, so in a sense, I guess it should be okay (as long as this is an optional secret), no?

> It can be any secret (generated from default-config or user created and referenced in CR) of auth.json format mounted to ~/.config}/containers/auth.json

Anyway, in such cases, I'd suggest perhaps deprecating this first and filing a docs issue rather than removing this right away (which is what you did in https://github.com/redhat-developer/rhdh-operator/pull/1297/commits/71899c135199a748842be3babd70b33d99db881a).

PR #1379, opened 2025-07-16
ci: Include `release 1.7` in the E2E nightly checks

rm3l:
## Description
This PR includes the newly created `release-1.7` in our nightly checks.

And to make it easier to handle excluded paths in the upgrade workflow, it automatically determines which paths are not suitable for testing (those with the same source and destination branches, or those involving downgrades). This way, we won't need to manually maintain that exclusion list.

&mdash;

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

Examples of runs:
- Regular E2E tests (failing because I don't have a Sealights agent token in my fork): https://github.com/rm3l/redhat-developer-hub-operator/actions/runs/16316241398
- Upgrade checks: https://github.com/rm3l/redhat-developer-hub-operator/actions/runs/16316020768

### DIFF
--- a/.github/workflows/nightly-upgrade-test.yaml
+++ b/.github/workflows/nightly-upgrade-test.yaml
@@ -15,8 +15,10 @@ jobs:
         from_branch:
           - release-1.5
           - release-1.6
+          - release-1.7
         to_branch:
           - main
+          - release-1.7
           - release-1.6
           - release-1.5
         exclude:
@@ -26,6 +28,12 @@ jobs:
             to_branch: release-1.5
           - from_branch: release-1.6
             to_branch: release-1.6
+          - from_branch: release-1.7
+            to_branch: release-1.5
+          - from_branch: release-1.7
+            to_branch: release-1.6
+          - from_branch: release-1.7
+            to_branch: release-1.7
 
     name: 'E2E Upgrade: ${{ matrix.from_branch }} => ${{ matrix.to_branch }}'
     concurrency:
--- a/.github/workflows/nightly-upgrade-test.yaml
+++ b/.github/workflows/nightly-upgrade-test.yaml
@@ -21,19 +21,6 @@ jobs:
           - release-1.7
           - release-1.6
           - release-1.5
-        exclude:
-          - from_branch: release-1.5
-            to_branch: release-1.5
-          - from_branch: release-1.6
-            to_branch: release-1.5
-          - from_branch: release-1.6
-            to_branch: release-1.6
-          - from_branch: release-1.7
-            to_branch: release-1.5
-          - from_branch: release-1.7
-            to_branch: release-1.6
-          - from_branch: release-1.7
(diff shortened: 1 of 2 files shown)


rm3l:
> * Consider moving the repeated step-level `if:` checks into a single job-level `if` so you don’t have to duplicate the skip condition on every step.

Not easily doable IMO because this is a matrix job and the condition depends on the values in that matrix.

> * Extract the upgrade-path skip logic into a composite action or shared script to keep the workflow file cleaner and improve maintainability.

For future improvement, maybe.

.github/workflows/nightly-upgrade-test.yaml
```diff
@@ -34,16 +29,41 @@ jobs:
     env:
       CONTAINER_ENGINE: podman
     steps:
+      - name: Determine if upgrade path should be skipped
+        id: upgrade-path-checker
+        run: |
+          SHOULD_BE_SKIPPED_BECAUSE_SAME_BRANCH_OR_DOWNGRADE="false"
+          from_branch=${{ matrix.from_branch }}
+          to_branch=${{ matrix.to_branch }}
+          if [[ "${from_branch}" == "main" ]]; then
```
Fortune-Ndlovu, line 38: Right now, the upgrade-path-checker has some tricky behavior when it comes to the `main` branch.

If you start on the `main` branch and try to upgrade to an older branch like `release-1.x`, that’s a downgrade, and it should be skipped. But if you start on an older branch like `release-1.x` and upgrade to main (which is the latest), that’s a valid upgrade and it should run.

The problem is that the current script might skip this valid upgrade because it compares version numbers (release-x.y format), and main doesn’t match that pattern. To fix this, we should explicitly say:

If the `to_branch` is `main`, always allow it (since upgrading to main is valid).

If the `from_branch` is `main`, always skip (since downgrading from main is invalid).

```bash
if [[ "${from_branch}" == "${to_branch}" ]]; then
  # Same branch, skip
  SHOULD_BE_SKIPPED_BECAUSE_SAME_BRANCH_OR_DOWNGRADE="true"
elif [[ "${to_branch}" == "main" ]]; then
  # Always allow upgrades to main
  SHOULD_BE_SKIPPED_BECAUSE_SAME_BRANCH_OR_DOWNGRADE="false"
elif [[ "${from_branch}" == "main" ]]; then
  # main -> release-x is always a downgrade, skip
  SHOULD_BE_SKIPPED_BECAUSE_SAME_BRANCH_OR_DOWNGRADE="true"
else
  # Normal numeric downgrade check
  ...
fi

```
That way:

Upgrading from main to an older release is correctly skipped.

Upgrading from an older release to main is correctly allowed.

Release-to-release upgrades still work as before.
WDYT?

Fortune-Ndlovu (commented):
Looks good, I also wonder, if we should explicitly handle main as a special case to avoid skipping valid upgrades

PR #1486, opened 2025-08-14
Merge dynamic-plugins config

gazarenkov:
## Description
This PR:

- Makes merging default and CR defined dynamic-plugins config
- Provides some refactoring of CM/Secret based configurations

https://issues.redhat.com/browse/RHIDP-7882

## PR acceptance criteria

- [x] Tests
- [x] Documentation

## Summary by Sourcery

Merge and refactor dynamic-plugins configuration handling by merging default and CR-defined plugin settings, unifying external configuration contributions via a new interface, simplifying object validation signatures, and enhancing tests and documentation.

New Features:
- Merge default and custom dynamic-plugins configurations to allow partial overrides

Enhancements:
- Introduce ExternalConfigContributor interface and unify external config handling for runtime objects
- Refactor RuntimeObject.updateAndValidate signature and store BackstageModel references within objects

Documentation:
- Document new dynamic-plugins merge behavior in configuration guide

Tests:
- Add and update tests for dynamic plugin merging, config map/env, secret/files contributors, and model initialization pipeline


pkg/model/dynamic-plugins.go
```diff
@@ -203,6 +197,90 @@ func (p *DynamicPlugins) pluginsFromConfigMap() ([]DynaPlugin, error) {
...
+			pluginMap[plugin.Package] = existingPlugin
+		} else {
+			// If the plugin is not found in model, add it from spec
+			if !plugin.Disabled {
+				pluginMap[plugin.Package] = plugin
+			}
+		}
+	}
+	mergedPluginsConfig := modelPluginsConfig
+	mergedPluginsConfig.Plugins = make([]DynaPlugin, 0, len(pluginMap))
```
Fortune-Ndlovu, line 245: I believe this may create a shallow copy, which could lead to unintended mutations of the original config?
```suggestion
    mergedPluginsConfig := DynaPluginsConfig{
        Includes: make([]string, 0, len(includeSet)),
        Plugins:  make([]DynaPlugin, 0, len(pluginMap)),
    }
```

pkg/model/dynamic-plugins.go
```diff
@@ -203,6 +197,90 @@ func (p *DynamicPlugins) pluginsFromConfigMap() ([]DynaPlugin, error) {
...
+func (p *DynamicPlugins) mergeWith(specData string) (string, error) {
+
+	if p.ConfigMap == nil {
+		return "", fmt.Errorf("dynamic plugins ConfigMap is not set")
+	}
+	modelData := p.ConfigMap.Data[DynamicPluginsFile]
+	var modelPluginsConfig, specPluginsConfig DynaPluginsConfig
+	if modelData != "" {
+		if err := yaml.Unmarshal([]byte(modelData), &modelPluginsConfig); err != nil {
+			return "", fmt.Errorf("failed to unmarshal model ConfigMap data: %w", err)
```
Fortune-Ndlovu, line 209: Should we make this error more readable e.g.:
```suggestion
			return "", fmt.Errorf("failed to unmarshal model ConfigMap data for key '%s': %w", 
    DynamicPluginsFile, err)
```

pkg/model/dynamic-plugins.go
```diff
@@ -203,6 +197,90 @@ func (p *DynamicPlugins) pluginsFromConfigMap() ([]DynaPlugin, error) {
...
+	}
+	for _, plugin := range specPluginsConfig.Plugins {
+
+		if existingPlugin, found := pluginMap[plugin.Package]; found {
+			if plugin.PluginConfig != nil {
+				existingPlugin.PluginConfig = plugin.PluginConfig
+			}
+			if len(plugin.Dependencies) > 0 {
+				existingPlugin.Dependencies = plugin.Dependencies
+			}
```
Fortune-Ndlovu, line 231: Because this is a complete replacement not a merge. if the default plugin have dependencies and the spec adds more are the defaults not lost? should we be considering merging somehow  https://github.com/redhat-developer/rhdh-operator/blob/main/docs/configuration.md

pkg/model/dynamic-plugins_test.go
```diff
@@ -319,3 +350,112 @@ plugins:
 	assert.NotNil(t, deps)
 	assert.Equal(t, 0, len(deps)) // Ensure it returns an empty slice, not nil
 }
+
+func TestMergeDynamicPlugins(t *testing.T) {
```
Fortune-Ndlovu, line 354: The existing TestMergeDynamicPlugins() only tests the mergeWith() function in isolation. Since you changed the core behavior from "replace" to "merge", we may need an integration test that verifies the full InitObjects(), merge, and mount flow works correctly with both default and external configs.

pkg/model/dynamic-plugins.go
```diff
@@ -203,6 +197,90 @@ func (p *DynamicPlugins) pluginsFromConfigMap() ([]DynaPlugin, error) {
...
+			if plugin.Integrity != "" {
+				existingPlugin.Integrity = plugin.Integrity
+			}
+			existingPlugin.Disabled = plugin.Disabled
+			pluginMap[plugin.Package] = existingPlugin
+		} else {
+			// If the plugin is not found in model, add it from spec
+			if !plugin.Disabled {
+				pluginMap[plugin.Package] = plugin
```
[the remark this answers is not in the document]
gazarenkov: Yes, thinking about includes (another merging) we need to keep disabled plugins here and later, fixed.

pkg/model/dynamic-plugins.go
gazarenkov: No real needs, as only one known key expected (as documented).

pkg/model/dynamic-plugins.go
gazarenkov: Yes, we do not merge the fields deeply, none of them.

rm3l (commented):
Overall LGTM - only this small comment.

rm3l:
@gazarenkov Also, could you please rebase your branch and fix the conflicts reported?

pkg/model/dynamic-plugins.go
Fortune-Ndlovu: so currently both `mergedPluginsConfig` and `modelPluginsConfig`  share the same [`includes`](https://github.com/gazarenkov/rhdh-operator/blob/external-config/pkg/model/dynamic-plugins.go#L49) slice in memory. so during include merging this could modify the original [`modelPluginsConfig.Includes`](https://github.com/gazarenkov/rhdh-operator/blob/external-config/pkg/model/dynamic-plugins.go#L49) if multiple reconciliations happen simultaneously. But I think we are actually fine as we create a new slice immediately.

gazarenkov:
@rm3l fixed, please check

rm3l (approved):

PR #1567, opened 2025-08-31
doc: add backend_secret config to orchestrator

jenniferubah:
## Description
To enable notification for Orchestrator, this PR adds the backend secret token under the auth config in the orchestrator example. So the user is aware that the secret needs to be configured for Orchestrator.

- Relates to bug ticket: https://issues.redhat.com/browse/FLPATH-2627

## PR acceptance criteria

- [x] Documentation

### DIFF
--- a/examples/orchestrator.yaml
+++ b/examples/orchestrator.yaml
@@ -53,4 +53,4 @@ spec:
     dynamicPluginsConfigMapName: orchestrator-plugin
     extraEnvs:
       secrets:
-        - name: backstage-backend-auth-secret
+        - name: backstage-backend-auth-secret # secret that contains the BACKEND_SECRET key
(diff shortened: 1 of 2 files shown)


examples/orchestrator.yaml
```diff
@@ -44,3 +51,6 @@ spec:
       configMaps:
         - name: app-config-rhdh
     dynamicPluginsConfigMapName: orchestrator-plugin
+    extraEnvs:
+      secrets:
+        - name: backstage-backend-auth-secret # secret that contains the BACKEND_SECRET key
```
rm3l, line 56: To make this example usable OOTB, how about adding the expected Secret in this file as well? It shouldn't matter if it contains some dummy data, as this is just a full example.

examples/orchestrator.yaml
rm3l, line 1: I wonder if the Helm Chart should be updated accordingly as well: https://github.com/redhat-developer/rhdh-chart/blob/main/charts/backstage/README.md#enablement-of-notifications-plugin
Or is the issue only impacting the Operator because it doesn't inject a `BACKEND_SECRET` env var in the app container by default?
WDYT?

PR #1619, opened 2025-09-10
Initial Qodo configuration

zdrapela:

- Fixes #_issue_number_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation


.pr_agent.toml
```diff
@@ -0,0 +1,35 @@
...
+publish_description_as_comment = true
+publish_description_as_comment_persistent = true
+add_original_user_description = false
+
+[pr_code_suggestions]
+commitable_code_suggestions = false
+
+[config]
+ignore_pr_authors = ["renovate","openshift-cherrypick-robot","rhdh-bot"]
+ignore_pr_labels = ["do-not-merge/work-in-progress"]
```
rm3l, line 34: It might make sense even on WIP PRs, no? I think automated review tools can help the PR author get some initial feedback, even if the PR is still in progress..

.pr_agent.toml
rm3l, line 1: Similar to https://github.com/redhat-developer/rhdh-operator/blob/main/.github/workflows/renovate-checks.yaml#L24 for Renovate configuration files, do you know if there is an existing tool to validate this configuration file? If not, maybe a simple toml linter would be enough at minimum? (But this can be addressed in a separate issue/PR).

.pr_agent.toml
```diff
@@ -0,0 +1,35 @@
+[jira]
+jira_api_token = "${{ secrets.JIRA_API_TOKEN }}"
```
rm3l, line 2: Just curious.. Where does this secret live? In the Qodo settings, I guess?

.pr_agent.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[pr_reviewer] # /review #
+persistent_comment = true
+require_tests_review = false
+require_ticket_analysis_review = true
+enable_review_labels_security = true
+enable_review_labels_effort = true
+
+[pr_description] # /describe #
+enable_pr_diagram = true
```
rm3l, line 21: Maybe it's only me, but I find these diagrams not that helpful in most cases..

.pr_agent.toml
rm3l: sounds good. This behavior is better. Thanks. So can we remove that `jira_api_token` field from the configuration then?

PR #2060, opened 2026-01-02
chore: bump operator-sdk and related tools versions and fix the issues reported

rm3l:
## Description

Bumps the version of the operator-sdk, which requires updating the linter configuration (and fixing the issues reported).

- Fixes https://issues.redhat.com/browse/RHIDP-11243

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation


.golangci.yml
```diff
@@ -1,14 +1,62 @@
...
+    - errcheck
+    - govet
+    - ineffassign
+    - misspell
+    - nakedret
     - staticcheck
+    - unconvert
+    - unused
+    # TODO(asoro): Enable these linters from the operator-sdk scaffold once the
+    # existing violations are fixed. Tracked as tech debt.
```
rm3l, line 17: Tracked as follow-up tech debt issue in https://redhat.atlassian.net/browse/RHDHBUGS-3358

Makefile
```diff
@@ -128,7 +127,7 @@ SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
 # Source packages outside of tests
-PKGS := $(shell go list ./... | grep -vE 'github.com/redhat-developer/rhdh-operator/(tests/|api/v1alpha([1-2]))')
+PKGS := $(shell go list ./... | grep -vE 'github.com/redhat-developer/rhdh-operator/(tests/|api/v1alpha([1-9]+))')
```
Fortune-Ndlovu: The PKGS regex changed from `v1alpha([1-2])` to `v1alpha([1-9]+)`, which now excludes all v1alpha versions including the active ones (v1alpha3-5). No test files exist in those packages today so nothing is lost. Does it make sense to tighten it to `v1alpha([1-5])` so a future v1alpha6 with tests wouldn't get silently skipped?

config/manager/kustomization.yaml
```diff
@@ -3,3 +3,211 @@ kind: Kustomization
...
+  target:
+    kind: Deployment
+
+# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
+# crd/kustomization.yaml
+#- path: manager_webhook_patch.yaml
+#  target:
+#    kind: Deployment
+
+# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER' prefix.
```
Fortune-Ndlovu, line 36: Lines 36-213 are entirely commented-out cert-manager/webhook scaffold boilerplate. Could we trim this to just the active lines plus a short comment like # See operator-sdk v1.42.2 scaffold for cert-manager/webhook config? It would cut ~170 lines of noise. We can leave as is if it helps you with your upcoming network policy feature

config/manager/manager_metrics_patch.yaml
```diff
@@ -0,0 +1,4 @@
+# This patch adds the args to allow exposing the metrics endpoint using HTTPS
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: --metrics-bind-address=:8443
```
Fortune-Ndlovu, line 4: On `main`, [`deployment.yaml`](https://github.com/redhat-developer/rhdh-operator/blob/main/config/manager/deployment.yaml#L55-L56) passes both `--metrics-bind-address=:8443` and `--metrics-secure=true` explicitly. This PR removes both from `deployment.yaml` and the [patch](https://github.com/rm3l/redhat-developer-hub-operator/blob/rhidp-11243-operator-upgrade-operator-sdk-version/config/manager/manager_metrics_patch.yaml) only adds back `--metrics-bind-address`. The secure flag now relies on the [Go code default changing from `false` to `true`](https://github.com/rm3l/redhat-developer-hub-operator/blob/rhidp-11243-operator-upgrade-operator-sdk-version/cmd/main.go#L70).

Could we add a second op here so the manifest stays explicit?

```yaml
- op: add
  path: /spec/template/spec/containers/0/args/-
  value: --metrics-secure=true
```

This way anyone reading the deployment knows metrics are secured without checking the Go source, and it protects against the default being changed back later.

Fortune-Ndlovu (commented):
Thanks for pushing the PR, it works as expected, added a few comments/nitpicks!

config/manager/kustomization.yaml
rm3l: Agreed it looks noisy, but IIRC, this is the output of the operator-sdk scaffold boilerplate. It's meant to be uncommented later when enabling cert-manager or webhooks. And it will indeep help with the network policies implementation (though that would be for the Operator's network policies feature). So I'd prefer to keep it as-is commented for now.

config/manager/manager_metrics_patch.yaml
rm3l: Yeah, this is intentional as it actually matches the operator-sdk v1.42.2 scaffold, which also passes `--metrics-bind-address` and relies on the Go default for `--metrics-secure` (which is now defaulting to `true` (same as the scaffold): https://github.com/operator-framework/operator-sdk/blob/v1.42.2/testdata/go/v4/memcached-operator/config/default/manager_metrics_patch.yaml
I was aiming to stay aligned with upstream as much as we can. Adding this second op would be a deviation we'd have to maintain next time we bump operator-sdk.

rm3l: > it protects against the default being changed back later.

People can still change it back to `false` even in this second op, no? I guess we'll rely on code review to detect any change to the default value.

Makefile
rm3l: well, there are no test files in those packages and they are also ignored by Codecov. But to your point, I've tightened it to `[1-3]` to match only the unserved API versions. Done in [`5cd26b9` (this PR)](https://github.com/redhat-developer/rhdh-operator/pull/2060/commits/5cd26b9b1eed94d057ad724f4a0a06010155845b)

config/manager/manager_metrics_patch.yaml
Fortune-Ndlovu: agreed, thank you for the additional context!

Fortune-Ndlovu (approved):
/lgtm

PR #2186, opened 2026-01-26
feat:  Generate ImageStream manifest for each package and make operator aware during reconciliation

Fortune-Ndlovu:
## Description
Create nightly GH workflow that would generate an ImageStream manifest for each package that we support and Make Operator aware of ImageStreams manifest and auto-create them during reconciliation

- Fixes #_issue_number_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation


rm3l (commented):
FYI, we discussed another simpler approach: push to registry.access.redhat.com (instead of registry.redhat.io), which doesn't require auth.
This is the scope of RHIDP-11519.
So the ImageStreams approach may not be needed anymore.

PR #2293, opened 2026-02-09
avoid using pull_request_target in workflows

gazarenkov:
## Description
Avoid using pull_request_target in workflows in workflows

**For Maintainers:** Comment `/build-images` to trigger image builds
**For Contributors:** Ask a maintainer to run `/build-images`


.github/workflows/pr-container-build.yaml
```diff
@@ -1,152 +1,163 @@
...
-          for file in ${ALL_CHANGED_FILES}; do
-            echo "$file was changed"
-          done
+          script: |
+            const pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number
+            });
+            core.setOutput('sha', pr.data.head.sha);
```
rm3l: From the security gem, this seems to be vulnerable to TOCTOU attacks. If an attacker modifies the PR between the moment a trusted maintainer triggers the slash command and the moment the actions/github-script executes, `pr.data.head.sha` will return the SHA of the newly pushed, unreviewed malicious commit.

rm3l:
Also, could you resolve the conflicts reported?

.github/workflows/pr-container-build.yaml
```diff
@@ -1,152 +1,163 @@
...
+            2. **Bundle:** \`quay.io/rhdh-community/operator-bundle:${BASE_VERSION}-pr-${PR_NUMBER}-${SHORT_SHA}\`
+            3. **Catalog:** \`quay.io/rhdh-community/operator-catalog:${BASE_VERSION}-pr-${PR_NUMBER}-${SHORT_SHA}\`
+
+            Also available with PR number tag:
+            - \`quay.io/rhdh-community/operator:${BASE_VERSION}-pr-${PR_NUMBER}\`
+            - \`quay.io/rhdh-community/operator-bundle:${BASE_VERSION}-pr-${PR_NUMBER}\`
+            - \`quay.io/rhdh-community/operator-catalog:${BASE_VERSION}-pr-${PR_NUMBER}\`
+
+            <sub>Triggered by @${USER_LOGIN}</sub>`
+            });
```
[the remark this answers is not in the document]
gazarenkov: We could but not sure I understand the point, sorry :) 
My understanding of the ask is to be able to test the image **easily**, right?
What is the point in relying on vars and provide facilities to multiply the same image?
I think:
- One publicly available image is enough (if needed)
- We have short and clear instruction how to build and use it and no needs to make it more difficult IMO
- Multiplying the image is not ecologically friendly :)

.github/workflows/pr-container-build.yaml
Zaperex: It'll just be a bit more annoying for reviewers but that should work.

rm3l (approved):

rm3l:
/override "PR Publish"
/override "PR Bundle Diff"
/override "PR Validate"

Those were triggered on `pull_request_target`.

rm3l (approved):

rm3l:
/override "PR Publish"

Those were triggered on `pull_request_target`.

PR #2633, opened 2026-04-08
[1.9] restore v1alpha1 and v1alpha2 as unserved

gazarenkov:
## Description

In a case if the cluster still contain v1alpha1 and|or v1alpha2 backstage CR versions listed in CRD status as storedVersion (possible even if there are no CR with such versions) upgrade to 1.9 is not possible as they were removed according to the plan.

To avoid this situation "non-hacky" way versions were restored but marked as unserved to avoid theirs using in the future.

https://redhat.atlassian.net/browse/RHDHSUPP-349

- Install operator 1.1.x version 
- create Backstage CR (stored version v1alpha1)
- upgrade to 1.9.3 (or other containing this fix) - should not fail


nickboldt:
minor nitpick, 1.9.3 is already released so this would land in 1.9.4.

https://github.com/redhat-developer/rhdh-operator/pull/2632

bundle/rhdh/metadata/annotations.yaml
```diff
@@ -4,7 +4,7 @@ annotations:
   operators.operatorframework.io.bundle.manifests.v1: manifests/
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: rhdh
-  operators.operatorframework.io.bundle.channels.v1: fast,fast-${CI_X_VERSION}.${CI_Y_VERSION}
+  operators.operatorframework.io.bundle.channels.v1: fast,fast-1.9.4
```
nickboldt:
```suggestion
  operators.operatorframework.io.bundle.channels.v1: fast,fast-${CI_X_VERSION}.${CI_Y_VERSION}
```

We don't have z-stream channels. Only y-stream ones.

bundle/rhdh/bundle.Dockerfile
```diff
@@ -5,7 +5,7 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=rhdh
-LABEL operators.operatorframework.io.bundle.channels.v1=fast,fast-${CI_X_VERSION}.${CI_Y_VERSION}
+LABEL operators.operatorframework.io.bundle.channels.v1=fast,fast-1.9.4
```
nickboldt:
```suggestion
LABEL operators.operatorframework.io.bundle.channels.v1=fast,fast-${CI_X_VERSION}.${CI_Y_VERSION}
```

We don't have z-stream channels. Only y-stream ones.

Makefile
```diff
@@ -40,7 +40,8 @@ ifeq ($(PROFILE), rhdh)
 	# IMAGE_TAG_BASE ?= registry.redhat.io/rhdh/rhdh-rhel9-operator
 	IMAGE_TAG_BASE ?= quay.io/rhdh/rhdh-rhel9-operator
 	DEFAULT_CHANNEL ?= fast
-	CHANNELS ?= fast,fast-\$${CI_X_VERSION}.\$${CI_Y_VERSION}
+	CHANNELS ?= fast,fast-${VERSION}
```
nickboldt: We don't have z-stream channels. Only y-stream ones. 


```suggestion
	CHANNELS ?= fast,fast-\$${CI_X_VERSION}.\$${CI_Y_VERSION}
```

gazarenkov: >We don't have z-stream channels. Only y-stream ones.
Yes, I know, it is for testing (and for reducing warning messages)
Do you use it? 
Will roll it back if so.
(other places - generated from here)

nickboldt (approved):
works for me, but needs an expert to approve.

Makefile
```diff
@@ -321,28 +321,92 @@ bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metada
...
+	@if [ "$(CONTAINER_TOOL)" = "podman" ]; then \
+		$(CONTAINER_TOOL) manifest create $(BUNDLE_IMG) \
+			$$(for platform in $$(echo $(BUNDLE_PLATFORMS) | tr ',' ' '); do \
+				echo "containers-storage:$(BUNDLE_IMG)-$$(echo $$platform | tr '/' '-')"; \
+			done); \
+	else \
+		$(CONTAINER_TOOL) manifest create $(BUNDLE_IMG) \
+			$$(for platform in $$(echo $(BUNDLE_PLATFORMS) | tr ',' ' '); do \
+				echo "$(BUNDLE_IMG)-$$(echo $$platform | tr '/' '-')"; \
+			done); \
```
rm3l, line 349: This doesn't seem to work with docker:
```
❯ make bundle-build CONTAINER_TOOL=docker                                                                                                                                                      
Building multi-arch bundle image for platforms: linux/amd64,linux/arm64                                                                                                                        
Building for linux/amd64...
[...]
Building for linux/arm64...
[...]
Creating manifest list...
no such manifest: quay.io/rhdh/rhdh-rhel9-operator-bundle:1.9-linux-amd64
make: *** [Makefile:333: bundle-build] Error 1
```

It works with podman however.

That being said, I understand that this might have helped test this on Mac, but I'm wondering if if it wouldn't be better to handle the changes related to multi-arch builds as a separate issue/PR (maybe as part of https://issues.redhat.com/browse/RHDHPLAN-1028 ) to focus here only on the CRD upgrade issue?

rm3l (commented):
Tested the CRD upgrade via OLM, and it worked correctly. Small concern on the image builds below.

Makefile
gazarenkov: The problem is that it is not possible to build a catalog for amd64 on Mac w/o having available for opm (local) arm64 image.
...investigating the problem

gazarenkov: I propose to document Docker problem as known issue and go ahead. @rm3l WDYT?

rm3l: Sounds good, as discussed together. I'll file a tech debt issue for later.

rm3l (approved):

Makefile
rm3l: https://redhat.atlassian.net/browse/RHDHBUGS-2968, fyi

PR #2661, opened 2026-04-16
Revert "[release-1.8] feat: bumping orchestrator plugins to 1.8.10 (#…

lholmquist:
…2646)"

This reverts commit ae1c02603bcb2e817a132c3a0b2cc1201679678f.

There will be another orchestrator plugin release which will most likely be 1.8.11 and that one should be used instead.

- Fixes #_issue_number_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/bundle/backstage.io/manifests/backstage-operator.clusterserviceversion.yaml
+++ b/bundle/backstage.io/manifests/backstage-operator.clusterserviceversion.yaml
@@ -45,7 +45,7 @@ metadata:
           }
         }
       ]
-    createdAt: "2026-04-14T18:37:58Z"
+    createdAt: "2026-03-30T15:48:46Z"
     description: Backstage Operator
     operators.operatorframework.io/builder: operator-sdk-v1.37.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v4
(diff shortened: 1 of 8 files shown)


rm3l:
> There will be another orchestrator plugin release which will most likely be 1.8.11 and that one should be used instead.

@lhomquist Are you reverting because the current 1.8.10 orchestrator plugins are broken? If not (and assuming that our 1.8.6 is expected to bump the plugins to 1.8.11 anyway), I'd suggest we just leave them as is on 1.8.10 and just update to 1.8.11 when that version is available. But if 1.8.10 is broken, fine to revert then. WDYT?

lholmquist:
No,  not broken,  just not the version that will be part of 1.8.6.  If you are ok with that, then feel free to close.  i also opened PRs to revert for both the rhdh-local and rhdh-charts,  i can close those two if you feel like we should close this one

rm3l:
> No, not broken, just not the version that will be part of 1.8.6. If you are ok with that, then feel free to close. i also opened PRs to revert for both the rhdh-local and rhdh-charts, i can close those two if you feel like we should close this one

In that case, let's close this and create new PRs when 1.8.11 is available. I'll close the other ones as well.

/close

PR #2719, opened 2026-04-23
[RHIDP-13101] Update lightspeed plugins and images

Jdubrick:
## Description
- Updates Lightspeed Core image tag for Lightspeed
- Updates Lightspeed plugin versions

https://redhat.atlassian.net/browse/RHIDP-13101

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/config/profile/rhdh/default-config/deployment.yaml
+++ b/config/profile/rhdh/default-config/deployment.yaml
@@ -67,7 +67,8 @@ spec:
             - name: NPM_CONFIG_USERCONFIG
               value: /opt/app-root/src/.npmrc.dynamic-plugins/.npmrc
             - name: MAX_ENTRY_SIZE
-              value: "30000000"
+              value: "40000000"
+            # CATALOG_INDEX_IMAGE will be replaced by the value of the `RELATED_IMAGE_catalog_index` env var, if set
             - name: CATALOG_INDEX_IMAGE
               value: "quay.io/rhdh/plugin-catalog-index:1.9"
             - name: CATALOG_ENTITIES_EXTRACT_DIR
--- a/bundle/rhdh/manifests/backstage-operator.clusterserviceversion.yaml
+++ b/bundle/rhdh/manifests/backstage-operator.clusterserviceversion.yaml
@@ -29,7 +29,7 @@ metadata:
     categories: Developer Tools
     certified: "true"
     containerImage: registry.redhat.io/rhdh/rhdh-rhel9-operator:1.10
-    createdAt: "2026-04-26T17:54:47Z"
+    createdAt: "2026-04-27T13:30:43Z"
     description: Red Hat Developer Hub is a Red Hat supported version of Backstage.
       It comes with pre-built plug-ins and configuration settings, supports use of
       an external database, and can help streamline the process of setting up a self-managed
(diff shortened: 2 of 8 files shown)


config/profile/rhdh/default-config/deployment.yaml
```diff
@@ -67,7 +67,8 @@ spec:
             - name: NPM_CONFIG_USERCONFIG
               value: /opt/app-root/src/.npmrc.dynamic-plugins/.npmrc
             - name: MAX_ENTRY_SIZE
-              value: "30000000"
+              value: "40000000"
```
rm3l, line 70: Would it work if this env var is overridden in each flavour?
- `config/profile/rhdh/default-config/flavours/lightspeed/deployment.yaml`
- `config/profile/rhdh/default-config/flavours/orchestrator/deployment.yaml`

Jdubrick: My thought process was that if it is in the main deployment (afaik there isn't flavour specific) then if both orchestrator and lightspeed are present it covers both cases, since orchestrator needed 30MB it is safe to take the larger value Lightspeed needs at 40MB.

Wdyt?

gazarenkov: @rm3l Let's not set it up for orchestrator if it requires less as flavours order is not guaranteed.

@Jdubrick We can do it in deployment.yaml but I'd probably recommend to define it in configmap-envs.yaml (as discussed in Slack) for better visibility.

Jdubrick: In that case, would Orchestrator need its own copy of that for its flavour? Since if we drop from `deployment.yaml` then Orchestrator won't have any area to configure it

rm3l: > Let's not set it up for orchestrator if it requires less as flavours order is not guaranteed.

Ah good catch, the order might indeed be an issue.

> would Orchestrator need its own copy of that for its flavour? Since if we drop from deployment.yaml then Orchestrator won't have any area to configure it

@Jdubrick I believe the point is to leave the default value to `30000000` in `config/profile/rhdh/default-config/deployment.yaml` (and that will cover the Orchestrator) and increase it to `40000000` in `config/profile/rhdh/default-config/flavours/lightspeed/configmap-envs.yaml`

Jdubrick: @rm3l @gazarenkov I made the changes locally and have been testing them out. I'm finding if I keep the `30000000` in the `deployment.yaml` and then have a `configmap-envs.yaml` in `/flavours/lightspeed` that overrides it, it fails to override it because of the precedence in Kube. The `env` in the `deployment.yaml` that is set to `30MB` always wins over the `envFrom` from the override in `/flavours/lightspeed`. 

What do you think? Obviously the easiest is to just have it in `deployment.yaml` but I am open to hearing your thoughts

gazarenkov: Ah, interesting... Indeed we did not test the case when it is defined in both deployment and configmap-envs...
Yes, please back it to deployment.yaml for now, will see then.

Jdubrick: It is in the `deployment.yaml` currently fyi @gazarenkov @rm3l

rm3l (approved):

PR #2733, opened 2026-04-29
do not delete disabled resources

gazarenkov:
## Description

It excludes potential deleting the resources created when certain feature is disabled (such as local db, route, monitor) 

https://redhat.atlassian.net/browse/RHDHBUGS-2781 

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

For example:
- Create empty CR (spec.database.enableLocalDb=true by default)
- Check DB StatefulSet created
- Modify CR with spec.database.enableLocalDb=false and apply
- Check DB StatefulSet is still there


rm3l (commented):
Hmm.., it might be confusing to see that some resources are still there when the corresponding feature is purposely toggled from enabled to disabled. For example, I first create a Backstage CR on OCP (so with a Route created by default), then I set `spec.application.route.enabled: false` in the CR, and now the Route object is still present.
Maybe the Operator could add some uniquely identifiable labels on the resources it creates and manages, and only try to delete those to avoid conflict with user-created resources?
Or (thinking out loud) maybe just refuse to reconcile if there are conflicting resources that it does not manage, and thus avoid touching those?

gazarenkov:
> Hmm.., it might be confusing to see that some resources are still there when the corresponding feature is purposely toggled from enabled to disabled. For example, I first create a Backstage CR on OCP (so with a Route created by default), then I set `spec.application.route.enabled: false` in the CR, and now the Route object is still present. Maybe the Operator could add some uniquely identifiable labels on the resources it creates and manages, and only try to delete those to avoid conflict with user-created resources? Or (thinking out loud) maybe just refuse to reconcile if there are conflicting resources that it does not manage, and thus avoid touching those?

Let's just document it for clarity and keep simple until we really need to make it complex.
Those resources, with specific names, are "special", operator manages it if certain feature is enabled and ignores otherwise.

rm3l:
> > Hmm.., it might be confusing to see that some resources are still there when the corresponding feature is purposely toggled from enabled to disabled. For example, I first create a Backstage CR on OCP (so with a Route created by default), then I set `spec.application.route.enabled: false` in the CR, and now the Route object is still present. Maybe the Operator could add some uniquely identifiable labels on the resources it creates and manages, and only try to delete those to avoid conflict with user-created resources? Or (thinking out loud) maybe just refuse to reconcile if there are conflicting resources that it does not manage, and thus avoid touching those?
> 
> Let's just document it for clarity and keep simple until we really need to make it complex. Those resources, with specific names, are "special", operator manages it if certain feature is enabled and ignores otherwise.

This would be kinda breaking change then, I guess.
But overall, I think that even with documentation, just leaving them as is might end up creating more confusion with orphaned resources that might still consume cluster resources. The example with the Route would be particularly confusing IMO as the route is still active (can accept external traffic) but is no longer managed by the Operator.
Since the operator already sets owner reference on the resources it creates, would it make sense to "just" check with `metav1.IsControlledBy` before deleting? I feel like this would be small targeted changed compared to the current PR which, IMHO, would be a behavioral change that might end up causing more confusion (in 1.10).

rm3l:
> > > Hmm.., it might be confusing to see that some resources are still there when the corresponding feature is purposely toggled from enabled to disabled. For example, I first create a Backstage CR on OCP (so with a Route created by default), then I set `spec.application.route.enabled: false` in the CR, and now the Route object is still present. Maybe the Operator could add some uniquely identifiable labels on the resources it creates and manages, and only try to delete those to avoid conflict with user-created resources? Or (thinking out loud) maybe just refuse to reconcile if there are conflicting resources that it does not manage, and thus avoid touching those?
> > 
> > 
> > Let's just document it for clarity and keep simple until we really need to make it complex. Those resources, with specific names, are "special", operator manages it if certain feature is enabled and ignores otherwise.
> 
> This would be kinda breaking change then, I guess. But overall, I think that even with documentation, just leaving them as is might end up creating more confusion with orphaned resources that might still consume cluster resources. The example with the Route would be particularly confusing IMO as the route is still active (can accept external traffic) but is no longer managed by the Operator. Since the operator already sets owner reference on the resources it creates, would it make sense to "just" check with `metav1.IsControlledBy` before deleting? I feel like this would be small targeted changed compared to the current PR which, IMHO, would be a behavioral change that might end up causing more confusion (in 1.10).

As discussed today, let's document this here (and also consider mentioning this in the official product docs).

rm3l (approved):
Getting this merged (so we can have it in time for 1.10) and I'll create a follow up issue to clarify the behavior in the docs (can be done in 1.10.z or 2.y)
EDIT: Created https://redhat.atlassian.net/browse/RHIDP-13670

PR #2755, opened 2026-04-30
(chore): update lightspeed plugin and configs

Jdubrick:
## Description

- Pulls lightspeed changes from upstream
- Updates lightspeed plugin to `2.6.0`

**Note:** The change to the `host` for lightspeed core is part of a security hardening to prevent access to the container from outside the RHDH Pod

## Which issue(s) does this PR fix or relate to

- Fixes #_issue_number_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

PR #2756, opened 2026-05-01
(chore): update lightspeed to use inherit, update catalog index to 1.10 image

Jdubrick:
## Description
- Updates Lightspeed plugins to use `{{inherit}}`
- Bumps catalog index image to `1.10`

https://redhat.atlassian.net/browse/RHIDP-13354

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/config/profile/rhdh/default-config/flavours/lightspeed/dynamic-plugins.yaml
+++ b/config/profile/rhdh/default-config/flavours/lightspeed/dynamic-plugins.yaml
@@ -6,7 +6,7 @@ data:
   dynamic-plugins.yaml: |
     plugins:
       # Lightspeed Plugins
-      - package: oci://ghcr.io/redhat-developer/rhdh-plugin-export-overlays/red-hat-developer-hub-backstage-plugin-lightspeed:bs_1.49.4__2.2.1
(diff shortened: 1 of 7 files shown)


Jdubrick:
Will cause a merge conflict with plugin change in https://github.com/redhat-developer/rhdh-operator/pull/2755, I will address it in this PR or that one whatever is done first

config/profile/rhdh/default-config/flavours/lightspeed/dynamic-plugins.yaml
```diff
@@ -33,5 +33,5 @@ data:
                     config:
                       id: lightspeed
                       priority: 100
```
rm3l, line 35: Can the whole `pluginConfig` field be removed?
With `{{inherit}}`, we should ideally be able to inherit the `pluginConfig` from the DPDY as well (unless there is a need to use a specific configuration here?)

Jdubrick: Marking this convo as resolved as we spoke in Slack, the pluginConfig in the DPDY is incorrectly formatted so we can't rely on `inherit` for the pluginConfig until that is fixed. This will be resolved in a follow up issue.

rm3l (approved):

PR #2779, opened 2026-05-05
(fix): add annotation for skipping reconciling for resources

Jdubrick:
## Description
- In the RHDH Helm chart, users of Lightspeed are able to provide their own copies of the configuration files, this is important because for some things (i.e. using postgres instead of sqlite, adding MCP servers) you need to edit the `lightspeed-stack.yaml` config file.
- The Lightspeed flavour had these config files managed by the Operator, and were being overwritten on reconciliation. This change adds a quick fix for that by introducing an annotation to skips the reconciliation if users want it to.
- The resources will still be managed by the Operator in terms of deletion/cleanup, and this will be reflected in the Lightspeed RHDH documentation, and a blurb has been added to `lightspeed.md` in this PR.
 

- Fixes #_issue_number_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/docs/lightspeed.md
+++ b/docs/lightspeed.md
@@ -91,6 +91,36 @@ spec:
 
 Ensure the secret contains the necessary authentication keys for AI service access.
 
+#### Preserving Manual Changes To Generated Configs
+
+After the first deployment, you can preserve manual changes to operator-created Lightspeed ConfigMaps by annotating the live object with `rhdh.redhat.com/prevent-overwrite`.
+
+This is useful when you want the operator to create the initial Lightspeed resources for you, and then stop overwriting later edits that you make directly in the cluster. Common examples include the generated `llama-stack-config`, `lightspeed-stack-config`, and `rhdh-profile` ConfigMaps.
+
+Example:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: lightspeed-stack-config
+  annotations:
+    rhdh.redhat.com/prevent-overwrite: "true"
+```
+
+Accepted values:
+
+- `true` in any letter case enables the behavior, for example `true`, `True`, or `TRUE`
+- Any other value, including `false` or an empty string, is treated as unset and the operator will continue reconciling the resource
+
+Behavior notes:
+
+- The annotation is checked on the existing live resource after it has already been created by the operator
+- When set to `true`, the operator skips future apply updates for that resource so your manual changes are preserved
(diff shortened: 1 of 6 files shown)


docs/lightspeed.md
```diff
@@ -91,6 +91,36 @@ spec:
 
 Ensure the secret contains the necessary authentication keys for AI service access.
 
+#### Preserving Manual Changes To Generated Configs
+
+After the first deployment, you can preserve manual changes to operator-created Lightspeed ConfigMaps by annotating the live object with `rhdh.redhat.com/prevent-overwrite`.
```
gazarenkov, line 96: @Jdubrick By idea you should be able to redefine the file in your own ConfigMap/Secret and reference it in your CR instead? Are there any problem with that?
So far we consider default configuration (including flavours) as totally operator managed. Can you see the reason to change it (other than reducing the number of resources, which is not that important IMO)?

gazarenkov (changes requested):

docs/lightspeed.md
Jdubrick: Similar to the Secret and how we include it here you mean? https://github.com/redhat-developer/rhdh-operator/blob/main/examples/lightspeed.yaml

If so:

My main reason for adding the annotation is because some users will not need to update the file, and some will. By having it deployed by the Operator, and if you want to add changes and prevent overwrite you add the annotation, makes it less work for those that don't need changes. 

Another reason why is that if we expect users to include this configmap (like the secret) when they deploy their Backstage CR and they forget to add that CM, Lightspeed will fail to start and throw errors, unlike the secret where if its missing it will still start up.

What do you think? If it is the desire to have it moved to being handled like the secret, I can try and do that

gazarenkov: By idea , custom ConfigMap (same as the Secret, yes) prevails on defined in default|flavour config with the same container and path (see https://github.com/redhat-developer/rhdh-operator/blob/main/pkg/model/configmapfiles_test.go#L117 ).
So user can create a configMap ("my-configmap") for e.g. lightspeed-stack.yaml and refer it in the spec like:
```
    extraFiles:
      configMaps:
        - name: "my-configmap"
          mountPath: /app-root
          key: lightspeed-stack.yaml
          containers:
            - lightspeed-core
```
and it has to replace one from default config.

Jdubrick: yup @gazarenkov you are right, we can use that way my apologies I didn't realize we could do that. Problem solved :) I am going to close this PR

PR #2824, opened 2026-05-12
fix: honor init container order in `spec.deployment.patch` [RHDHBUGS-2900]

rm3l:
## Description

`spec.deployment.patch` did not allow users to control the ordering of init containers. New init containers would always be appended after existing ones (e.g., `install-dynamic-plugins` always ran first).

This PR switches the `kyaml` merge strategy so new list items are prepended before existing ones by default. Users who need a custom init container to run after an existing one can still anchor it by referencing the existing container by name first in the patch (this didn't work with the default merge strategy).

- Fixes https://redhat.atlassian.net/browse/RHDHBUGS-2900

## PR acceptance criteria

- [x] Tests
- [x] Documentation

Note that a two-pass merge is used here to work around https://github.com/kubernetes-sigs/kustomize/issues/6146,  where the prepend strategy silently breaks `$patch` directives. Once that issue is fixed in kyaml, we'll be able to remove this two-pass merge and only use the Prepend strategy.


pkg/model/deployment.go
```diff
@@ -227,7 +227,20 @@ func (b *BackstageDeployment) setDeployment(backstage api.Backstage) error {
 				return fmt.Errorf("can not marshal deployment object: %w", err)
 			}
 
-			merged, err := merge2.MergeStrings(string(conf.Raw), string(deplStr), false, kyaml.MergeOptions{})
+			// TODO(asoro): once https://github.com/kubernetes-sigs/kustomize/issues/6146 is resolved,
+			// remove this two-pass merge and use only ListPrepend.
+			//
+			// RHDHBUGS-2900: Two-pass merge to work around a kyaml bug where ListPrepend
+			// silently breaks $patch directives (e.g. $patch: replace).
```
gazarenkov: I could not reproduce this bug (tried with single MergeStrings and it worked with directives as well). What exactly is your config to reproduce?

pkg/model/deployment_test.go
```diff
@@ -140,23 +140,112 @@ spec:
 	assert.Equal(t, "java", model.backstageDeployment.deployable.GetObject().GetLabels()["mylabel"])
 	assert.Equal(t, "backstage", model.backstageDeployment.deployable.PodObjectMeta().GetLabels()["pod"])
 
-	// sidecar added
+	// sidecar prepended
 	assert.Equal(t, 2, len(model.backstageDeployment.podSpec().Containers))
-	assert.Equal(t, "sidecar", model.backstageDeployment.podSpec().Containers[1].Name)
-	assert.Equal(t, "my-image:1.0.0", model.backstageDeployment.podSpec().Containers[1].Image)
+	assert.Equal(t, "sidecar", model.backstageDeployment.podSpec().Containers[0].Name)
```
gazarenkov: This looks like breaking change for users relying on certain order (appended logic).
We should at least document it (however I do not think we have it documented anywhere, so probably just Release Notes) or even think about additional conf flag like:
`spec.deployment.mergeListMode: [append]|prepend`
?

pkg/model/deployment.go
rm3l: Sure, without this, one of the tests fails:

<details>
<summary>Details</summary>

```
[...]
------------------------------                                                                                                                                                                 
• [FAILED] [10.109 seconds]                                                                                                                                                                    
when create default rhdh [It] replaces dynamic-plugins-root volume                                                                                                                             
/home/asoro/work/projects/backstage/janus-idp/operator/integration_tests/rhdh-config_test.go:200                                                                                               
                                                                                                                                                                                               
  Timeline >>                                                                                                                                                                                  
  STEP: getting the Pod  @ 05/19/26 11:22:12.827                                                                                                                                               
  STEP: getting the Pod  @ 05/19/26 11:22:13.829                                                                                                                                               
  STEP: getting the Pod  @ 05/19/26 11:22:14.831                                                                                                                                               
  STEP: getting the Pod  @ 05/19/26 11:22:15.834                                                                                                                                               
  STEP: getting the Pod  @ 05/19/26 11:22:16.837                                                                                                                                               
  STEP: getting the Pod  @ 05/19/26 11:22:17.84                                                                                                                                                
  STEP: getting the Pod  @ 05/19/26 11:22:18.843                                                                                                                                               
  STEP: getting the Pod  @ 05/19/26 11:22:19.846                                                                                                                                               
  STEP: getting the Pod  @ 05/19/26 11:22:20.849                                                                                                                                               
  STEP: getting the Pod  @ 05/19/26 11:22:21.852                                                                                                                                               
  [FAILED] in [It] - /home/asoro/work/projects/backstage/janus-idp/operator/integration_tests/rhdh-config_test.go:236 @ 05/19/26 11:22:22.827                                                  
  << Timeline                                                                                                                                                                                  
                                                                                                                                                                                               
  [FAILED] Timed out after 10.000s.                                                                                                                                                            
  The function passed to Eventually failed at /home/asoro/work/projects/backstage/janus-idp/operator/integration_tests/rhdh-config_test.go:233 with:                                           
  Expected                                                                                                                                                                                     
      <*v1.EphemeralVolumeSource | 0xc00058a068>: {                                                                                                                                            
          VolumeClaimTemplate: {                                                                                                                                                               
              ObjectMeta: {                                                                                                                                                                    
                  Name: "",                                                                                                                                                                    
                  GenerateName: "",                                                                                                                                                            
                  Namespace: "",                                                                                                                                                               
                  SelfLink: "",                                                                                                                                                                
                  UID: "",                                                                                                                                                                     
                  ResourceVersion: "",                                                                                                                                                         
                  Generation: 0,                                                                                                                                                               
                  CreationTimestamp: {                                                                                                                                                         
                      Time: 0001-01-01T00:00:00Z,                                                                                                                                              
                  },                                                                                                                                                                           
                  DeletionTimestamp: nil,
                  DeletionGracePeriodSeconds: nil,
                  Labels: nil,
                  Annotations: nil,
                  OwnerReferences: nil,
                  Finalizers: nil,
                  ManagedFields: nil,
              },
              Spec: {
                  AccessModes: ["ReadWriteOnce"],
                  Selector: nil,
                  Resources: {
                      Limits: nil,
                      Requests: {
                          "storage": {
                              i: {value: 2147483648, scale: 0},
                              d: {Dec: nil},
                              s: "2Gi",
                              Format: "BinarySI",
                          },
                      },
                  },
                  VolumeName: "",
                  StorageClassName: nil,
                  VolumeMode: "Filesystem",
                  DataSource: nil,
                  DataSourceRef: nil,
                  VolumeAttributesClassName: nil,
              },
          },
      }
  to be nil
  In [It] at: /home/asoro/work/projects/backstage/janus-idp/operator/integration_tests/rhdh-config_test.go:236 @ 05/19/26 11:22:22.827
------------------------------
```

</details>

And I confirmed this behavior with the manifest I sent you on Slack:

```yaml
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: dynamic-plugins-root-pvc
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 2Gi

---
kind: Backstage
metadata:
  name: rhdh-2900-05
spec:
  deployment:
    patch:
      spec:
        replicas: 1
        template:
          spec:
            initContainers:
              - name: pre-init
                image: docker.io/busybox
                command: ["echo", "Hello from pre-init (should run BEFORE install-dynamic-plugins)"]
#              - name: init-rag-data
              - name: install-dynamic-plugins
              - name: post-init
                image: docker.io/busybox
                command: ["echo", "Hello from post-init (should run AFTER install-dynamic-plugins)"]
            volumes:
              - $patch: replace
                name: dynamic-plugins-root
                persistentVolumeClaim:
                  claimName: dynamic-plugins-root-pvc
```

After reconciliation, we can see that the dynamic-plugins-root volume is still ephemeral, and not replaced by the PVC:
```
[...]
  volumes:
  - ephemeral:
      volumeClaimTemplate:
        metadata: {}
        spec:
          accessModes:
          - ReadWriteOnce
          resources:
            requests:
              storage: 2Gi
          volumeMode: Filesystem
    name: dynamic-plugins-root
[...]
```

pkg/model/deployment_test.go
rm3l: Yeah, as this behavior is currently not documented anywhere, my plan was indeed to mention this in the release notes.
And I also thought of ultimately exposing the merge list mode, so users can have a choice. That can be the scope of a separate Feature. That being said, I guess `prepend` should already cover most cases, and users can still have full control by explicitly listing all containers, so not sure it is worth spending time exposing that ability tbh.

pkg/model/deployment.go
gazarenkov: Confirmed, patch replace does not work on array element level with prepend option (other directives seems to work).

pkg/model/deployment_test.go
gazarenkov: Do not you consider it as risky?

See, what we are trying to solve is quite a rare usecase and the way we solve it (double call) is pretty hacky.
I agree changing spec is not in time, but can we then consider an annotation for this purpose, like:
r.r.c/deployment-merge-list-mode: prepend which would cause first call and go usual way otherwise?

rm3l: > Do not you consider it as risky?

What concrete risks do you foresee?
What matters most IMO is the order of init containers. I believe the order of the other app containers should not matter. And, as stated in the docs, users still have full control over the ordering. Without this, no ordering possible at all, as raised in this customer issue.

> can we then consider an annotation for this purpose, like: r.r.c/deployment-merge-list-mode: prepend which would cause first call and go usual way otherwise?

Where would you put that annotation?
I agree with the two merges being hacky (well, hopefully temporarily until https://github.com/kubernetes-sigs/kustomize/issues/6146 is fixed, which I have tested a fix for in https://github.com/rm3l/kyaml-listprepend-bug/blob/fix/use-patched-kyaml/go.mod#L7 that would avoid this double merge).
Otherwise, with what you are suggesting, we would need to maintain two paths in the docs, no? One where ordering doesn't actually work, and another with controlled ordering if you set the annotation.

gazarenkov: I see the risk in hacky way of using library and in fact that this change is kinda breaking.

That's why I'd prefer to have a way to choose behaviour, defaulting to old one. This can be left even after fix (passing to single merge call)

Annotation in Backstage CR (it comes to setDeployment method I think)

rm3l: Okay, I understand. Let me add this..

rm3l: Made this opt-in in [`53bec37` (this PR)](https://github.com/redhat-developer/rhdh-operator/pull/2824/commits/53bec37f9eb39288a6eb9ee2e0c49bc1f272b721). PTAL.

docs/configuration.md
```diff
@@ -837,6 +837,62 @@ spec:
                   claimName: dynamic-plugins-root
 ```
 
+##### List ordering
+
+By default, new list items in the patch (containers, init containers, volumes, etc.) are **appended** after existing items. Items that match an existing entry by name are merged in-place and retain their original position — their placement in the patch does not reorder them in the result.
+
+If the ordering of list items matters (e.g., init containers, where execution order is [significant](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior)), you can opt in to **prepend** mode by setting the `rhdh.redhat.com/deployment-patch-list-merge-mode` annotation on the Backstage CR. In prepend mode, you can control the exact ordering by listing all items (including existing ones) explicitly in the patch in the desired order.
+
+> **Note:** The `$setElementOrder` directive from the Kubernetes Strategic Merge Patch specification is **not supported** by the [kyaml](https://github.com/kubernetes-sigs/kustomize/tree/master/kyaml) (kustomize) library used by this operator. Listing all items explicitly in the patch is the recommended way to control ordering.
```
gazarenkov: $deleteFromPrimitiveList and $retainKeys are not supported as well

gazarenkov (approved):

PR #2828, opened 2026-05-12
ci: upload unit test coverage to Codecov [RHIDP-13388]

rm3l:
## Description
Add codecov.yml configuration and upload steps to the PR and nightly workflows. Coverage checks are informational only and will not block merges.

- Fixes [RHIDP-13388](https://redhat.atlassian.net/browse/RHIDP-13388)

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/codecov-checks.yaml
+++ b/.github/workflows/codecov-checks.yaml
@@ -1,0 +1,34 @@
+name: Codecov config validator
+
+on:
+  push:
+    paths:
+      - 'codecov.yml'
+      - '.github/workflows/codecov-checks.yaml'
+  pull_request:
+    paths:
+      - 'codecov.yml'
+      - '.github/workflows/codecov-checks.yaml'
+
+jobs:
+  codecov_validate:
+    runs-on: ubuntu-latest
+    env:
+      CODECOV_CONFIG_FILE: codecov.yml
+      CODECOV_VALIDATE_URL: https://codecov.io/validate
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - name: Validate codecov.yml
+        run: |
+          response=$(curl -s -o /dev/null -w "%{http_code}" --data-binary @"${CODECOV_CONFIG_FILE}" "${CODECOV_VALIDATE_URL}")
+          if [ "$response" -eq 200 ]; then
+            echo "${CODECOV_CONFIG_FILE} is valid"
+            exit 0
+          fi
+          if [ "$response" -eq 400 ]; then
+            echo "${CODECOV_CONFIG_FILE} is invalid (HTTP $response)"
+            curl --data-binary @"${CODECOV_CONFIG_FILE}" "${CODECOV_VALIDATE_URL}"
+            exit 1
+          fi
+          echo "::warning ::Codecov validation endpoint returned HTTP $response. This may be a transient issue — please retry."
+          exit 1
--- a/.github/workflows/codecov-checks.yaml
+++ b/.github/workflows/codecov-checks.yaml
@@ -35,5 +35,5 @@ jobs:
             exit 1
           fi
           echo "::warning ::Codecov validation endpoint returned HTTP $http_code after retries. This may be a transient issue — please re-run the job."
-          # Don't fail the job on transient Codecov API issues
-          exit 0
+          # Fail earlier on transient Codecov API issues, so we can catch the issue and eventually retry
+          exit 1
--- a/.github/workflows/codecov-checks.yaml
+++ b/.github/workflows/codecov-checks.yaml
@@ -22,14 +22,18 @@ jobs:
         env:
           CURL_OPTS: "--connect-timeout 10 --max-time 30 --retry 3 --retry-all-errors"
         run: |
-          response=$(curl -s -o /dev/null -w "%{http_code}" ${CURL_OPTS} --data-binary @"${CODECOV_CONFIG_FILE}" "${CODECOV_VALIDATE_URL}")
-          if [ "$response" -eq 200 ]; then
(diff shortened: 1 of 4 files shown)


codecov.yml
```diff
@@ -0,0 +1,39 @@
...
+comment:
+  layout: "reach,diff,flags,files"
+  behavior: default
+  require_changes: true
+  require_base: false
+  require_head: true
+
+flag_management:
+  individual_flags:
+    - name: unittests
```
gustavolira, line 28: `nightly` is used as a flag in the nightly workflow (`flags: unittests,nightly`) but only `unittests` is declared here. Without its own entry, `nightly` won't get `carryforward: true` -- so if a nightly run fails before the upload step, coverage data for that flag goes stale. Worth adding a second entry, or dropping `nightly` from the flags list if it's just for UI filtering.

.github/workflows/codecov-checks.yaml
```diff
@@ -0,0 +1,35 @@
...
+          if [ "$response" -eq 200 ]; then
+            echo "${CODECOV_CONFIG_FILE} is valid"
+            exit 0
+          fi
+          if [ "$response" -eq 400 ]; then
+            echo "${CODECOV_CONFIG_FILE} is invalid (HTTP $response)"
+            curl ${CURL_OPTS} --data-binary @"${CODECOV_CONFIG_FILE}" "${CODECOV_VALIDATE_URL}"
+            exit 1
+          fi
+          echo "::warning ::Codecov validation endpoint returned HTTP $response after retries. This may be a transient issue — please re-run the job."
```
gustavolira: nit: After this line, the script exits with whatever `echo` returns (0), so unexpected HTTP responses silently pass the job. That's probably the right call (don't block on Codecov API issues), but an explicit `exit 0` would make the intent clear. Without it, it reads like someone forgot the exit statement.

.github/workflows/codecov-checks.yaml
```diff
@@ -0,0 +1,35 @@
...
+          CURL_OPTS: "--connect-timeout 10 --max-time 30 --retry 3 --retry-all-errors"
+        run: |
+          response=$(curl -s -o /dev/null -w "%{http_code}" ${CURL_OPTS} --data-binary @"${CODECOV_CONFIG_FILE}" "${CODECOV_VALIDATE_URL}")
+          if [ "$response" -eq 200 ]; then
+            echo "${CODECOV_CONFIG_FILE} is valid"
+            exit 0
+          fi
+          if [ "$response" -eq 400 ]; then
+            echo "${CODECOV_CONFIG_FILE} is invalid (HTTP $response)"
+            curl ${CURL_OPTS} --data-binary @"${CODECOV_CONFIG_FILE}" "${CODECOV_VALIDATE_URL}"
```
gustavolira: nit: This is a second network request to the same endpoint just to show the error body -- the first call already hit it but discarded the body with `-o /dev/null`. The two calls could return different results. One way to capture both in a single call:

```bash
body=$(curl -s -w "\n%{http_code}" ${CURL_OPTS} --data-binary @"${CODECOV_CONFIG_FILE}" "${CODECOV_VALIDATE_URL}")
http_code=$(echo "$body" | tail -1)
body=$(echo "$body" | sed '$d')
```

codecov.yml
```diff
@@ -0,0 +1,39 @@
...
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 1%
+        if_ci_failed: error
+        informational: true
+    patch:
+      default:
+        target: 80%
```
gustavolira: nit: 80% patch target is ambitious for an initial rollout. Since everything is `informational: true` it won't block merges, but noisy "failing" checks can train people to ignore them. Consider starting with `target: auto` (same as project) and tightening once there's a baseline.

codecov.yml
rm3l: Good point. Added in [`ae51116` (this PR)](https://github.com/redhat-developer/rhdh-operator/pull/2828/commits/ae51116aaf0c1c6cf292b09e312433a2590b7325)

.github/workflows/codecov-checks.yaml
rm3l: Added in [`ae51116` (this PR)](https://github.com/redhat-developer/rhdh-operator/pull/2828/commits/ae51116aaf0c1c6cf292b09e312433a2590b7325)

.github/workflows/codecov-checks.yaml
rm3l: Addressed in [`ae51116` (this PR)](https://github.com/redhat-developer/rhdh-operator/pull/2828/commits/ae51116aaf0c1c6cf292b09e312433a2590b7325)

codecov.yml
rm3l: Addressed in [`ae51116` (this PR)](https://github.com/redhat-developer/rhdh-operator/pull/2828/commits/ae51116aaf0c1c6cf292b09e312433a2590b7325)

.github/workflows/codecov-checks.yaml
```diff
@@ -0,0 +1,39 @@
...
+            echo "${CODECOV_CONFIG_FILE} is valid"
+            exit 0
+          fi
+          if [ "$http_code" -eq 400 ]; then
+            echo "${CODECOV_CONFIG_FILE} is invalid (HTTP $http_code)"
+            exit 1
+          fi
+          echo "::warning ::Codecov validation endpoint returned HTTP $http_code after retries. This may be a transient issue — please re-run the job."
+          # Don't fail the job on transient Codecov API issues
+          exit 0
```
rm3l:
```suggestion
          # Fail earlier on transient Codecov API issues, so we can catch the issue and eventually retry
          exit 1
```

gustavolira (approved):

PR #2870, opened 2026-05-20
perf(scripts): speed up RHDH operator install using install-rhdh-catalog-source.sh (~30 min → ~2 min)

subhashkhileri:
JIRA : https://redhat.atlassian.net/browse/RHIDP-13676
## Summary

- **Skips slow `skopeo inspect`** (~42s/bundle) — attempts the copy directly; failed copies (~3s) are faster than successful inspects, so images that don't exist on quay are handled with minimal overhead
- **Parallel bundle processing** with `MAX_PARALLEL=10` (env-overridable)
- **Parallel `opm render` + registry setup** — these two phases are independent, so they now run concurrently; bundle processing waits until both are done
- **Atomic render.yaml updates** — each worker writes a per-bundle `.sed` file (keyed by digest, no locking needed); all replacements are applied in a single `sed -f` pass after all workers complete
- **`--ignore-not-found`** replaces the check-then-delete secret pattern
- **Force OLM re-index** by deleting the CatalogSource before recreating it — without this, OLM skips re-indexing when the tag is unchanged but the image digest changed (rebuilt IIB)
- **Fail loudly** if any bundle fails to process (was: silent error log + continue)

Assisted-by: Claude Code

### DIFF
--- a/.rhdh/scripts/install-rhdh-catalog-source.sh
+++ b/.rhdh/scripts/install-rhdh-catalog-source.sh
@@ -151,9 +151,7 @@ function k8s_check_bundle_manifest_default_config() {
   echo "ok"
 }
 
-# Process a single bundle: pull, rewrite refs, repack, push.
-# Skips the expensive skopeo inspect — attempts the copy directly and handles failure.
-# Writes sed replacement commands to a file for the caller to apply to render.yaml after all bundles complete.
+# Writes sed replacement commands to sed_commands_dir for the caller to batch-apply after all bundles complete.
 function process_bundle() {
   set -euo pipefail
 
@@ -226,7 +224,7 @@ function update_refs_in_iib_bundles() {
     bundleImg="${bundleImg/registry-proxy.engineering.redhat.com\/rh-osbs\/rhdh-/quay.io\/rhdh\/}"
     debugf "bundle #${bundle_count}/${total_bundles}: $originalBundleImg => $bundleImg" >&2
 
-    # Throttle: wait until a slot opens (portable, no wait -n needed)
+    # Portable alternative to `wait -n` (not available in all bash versions)
     while true; do
       local running=0
       for pid in ${pids[@]+"${pids[@]}"}; do
@@ -244,7 +242,6 @@ function update_refs_in_iib_bundles() {
     pids+=($!)
   done
 
-  # Wait for all remaining background bundle jobs
   local failed=0
   for pid in ${pids[@]+"${pids[@]}"}; do
     if ! wait "$pid"; then
@@ -256,7 +253,6 @@ function update_refs_in_iib_bundles() {
     return 1
   fi
 
-  # Apply all sed replacements to render.yaml
   local sed_files
   sed_files=$(find "$sed_commands_dir" -name '*.sed' 2>/dev/null)
   if [[ -n "$sed_files" ]]; then
@@ -285,7 +281,7 @@ function ocp_install() {
 
   set -euo pipefail
 
-  # Run opm render and registry setup in parallel — they're independent.
+  # render_iib is independent of registry setup below, so run concurrently.
   render_iib >&2 &
   local render_pid=$!
 
@@ -317,7 +313,6 @@ function ocp_install() {
   oc policy add-role-to-user system:image-puller system:serviceaccount:openshift-marketplace:default -n openshift-marketplace >&2 || true
   oc policy add-role-to-user system:image-puller system:serviceaccount:rhdh-operator:default -n rhdh-operator >&2 || true
 
-  # Wait for opm render to finish before processing bundles
   if ! wait "$render_pid"; then
     errorf "opm render failed" >&2
     return 1
--- a/.rhdh/scripts/install-rhdh-catalog-source.sh
+++ b/.rhdh/scripts/install-rhdh-catalog-source.sh
@@ -14,7 +14,7 @@ NAMESPACE_SUBSCRIPTION="rhdh-operator"
 OLM_CHANNEL="fast"
 UPSTREAM_IIB_OVERRIDE=""
 INSTALL_PLAN_APPROVAL="Automatic"
-MAX_PARALLEL="${MAX_PARALLEL:-10}" # processes bundles in parallel.
+MAX_PARALLEL="${MAX_PARALLEL:-10}" # max concurrent bundle workers in process_bundle
 
 function logf() {
   set -euo pipefail
--- a/.rhdh/scripts/install-rhdh-catalog-source.sh
+++ b/.rhdh/scripts/install-rhdh-catalog-source.sh
@@ -626,7 +626,8 @@ pushd "${TMPDIR}" > /dev/null
 debugf ">>> WORKING DIR: $TMPDIR <<<"
 
 # shellcheck disable=SC2064
-trap "rm -fr $TMPDIR || true; kill 0" EXIT INT TERM
+trap "rm -fr '$TMPDIR' || true; jobs -p | xargs -r kill 2>/dev/null; wait 2>/dev/null" EXIT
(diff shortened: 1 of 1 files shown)


.rhdh/scripts/install-rhdh-catalog-source.sh
```diff
@@ -150,56 +151,120 @@ function k8s_check_bundle_manifest_default_config() {
...
+
+    # Portable alternative to `wait -n` (not available in all bash versions)
+    while true; do
+      local running=0
+      for pid in ${pids[@]+"${pids[@]}"}; do
+        if kill -0 "$pid" 2>/dev/null; then
+          running=$((running + 1))
+        fi
       done
+      if [[ $running -lt $MAX_PARALLEL ]]; then
```
rm3l, line 239: I feel like we should validate that `MAX_PARALLEL` > 0. If I set it to 0, the script never enters this and hangs indefinitely.

rm3l, line 239: I'd also suggest validating that `MAX_PARALLEL` is a number, otherwise: 
```
[INFO] Processing 42 bundles (max zzz parallel)...
[DEBUG] bundle #1/42: registry.redhat.io/rhdh/rhdh-operator-bundle@sha256:2abaeacfa8fd744579e44e4b320086a8678094dd92eb24825c05f43617384529 => quay.io/rhdh/rhdh-operator-bundle@sha256:2abaeacfa8fd744579e44e4b320086a8678094dd92eb24825c05f43617384529
.rhdh/scripts/install-rhdh-catalog-source.sh: line 235: zzz: unbound variable
```

.rhdh/scripts/install-rhdh-catalog-source.sh
```diff
@@ -150,56 +151,120 @@ function k8s_check_bundle_manifest_default_config() {
...
 
-      # repack the image with the changes
-      debugf "Repacking image ./bundles/${digest}/src => ./bundles/${digest}/unpacked..."
-      umoci repack --image "./bundles/${digest}/src:latest" "./bundles/${digest}/unpacked"
-
-      # Push the bundle to the internal cluster registry
-      debugf "Pushing updated image: ./bundles/${digest}/src => ${newBundleImage}..."
-      skopeo copy --dest-tls-verify=false "oci:./bundles/${digest}/src:latest" "docker://${newBundleImage}"
+    process_bundle "$bundleImg" "$originalBundleImg" "$digest" "$my_registry" "$internal_registry_url" "$sed_commands_dir" "$bundle_count" &
+    pids+=($!)
```
rm3l, line 246: Are these correctly cleaned up? When I run the script, I'm left with zombie processes after the execution, like so (after 3 runs):
```
    PID    PPID CMD                         %CPU %MEM
  50275   48801 /usr/bin/zsh                 0.0  0.0
 185315       1 /bin/bash .rhdh/scripts/ins  0.0  0.0
 185317  185315 kubectl port-forward servic  0.0  0.0
 202911       1 /bin/bash .rhdh/scripts/ins  0.0  0.0
 202913  202911 kubectl port-forward servic  0.0  0.0
 212945       1 /bin/bash .rhdh/scripts/ins  0.0  0.0
 212947  212945 kubectl port-forward servic  0.0  0.0
 427780       1 kubectl port-forward servic  0.0  0.0
```

Also, what happens if the script is interrupted in the middle of this background execution? I guess the script should make sure to kill the process group on exit.

subhashkhileri: So this was actually a pre-existing bug the script had 3 separate `trap EXIT` calls (TMPDIR cleanup, port-forward kill, kaniko logs kill) that kept overwriting each other, meaning only the last one ever ran. Replaced all of them with a single `trap "rm -fr $TMPDIR || true; kill 0" EXIT INT TERM` that cleans up TMPDIR first, then kills the entire process group.

rm3l: Ah right, indeed, confirmed the same behavior on the main branch.
But after testing your new changes, the script always exits with an error, even when everything was perfectly fine, which makes me suspect a potential `trap` issue.:
```
[...]
catalogsource.operators.coreos.com/rhdh-fast created
[DEBUG] Creating OperatorGroup to allow all-namespaces operators to be installed
operatorgroup.operators.coreos.com/rhdh-operator-group unchanged
subscription.operators.coreos.com/rhdh unchanged
[...]

[1]    1834200 segmentation fault (core dumped)  .rhdh/scripts/install-rhdh-catalog-source.sh --latest --install-operator rhdh
```

Which is confirmed by Claude and [Qodo](https://github.com/redhat-developer/rhdh-operator/pull/2870#discussion_r3296943202):

```
Critical: kill 0 kills the parent process group (line 629)                                               
                                                                                                           
  trap "rm -fr $TMPDIR || true; kill 0" EXIT INT TERM                                                      
                                                                                                           
  kill 0 sends SIGTERM to every process in the current process group, not just children of this script. In
  a non-interactive shell (CI pipelines, scripts calling this script), the calling process shares the same
  process group. So:

  # In a CI pipeline or wrapper script (non-interactive shell):
  ./install-rhdh-catalog-source.sh --latest --install-operator rhdh
  ./run-tests.sh   # never reached — killed by the trap above

  When install-rhdh-catalog-source.sh exits (even successfully), its EXIT trap fires kill 0, which sends
  SIGTERM to the entire process group — including the parent shell running the pipeline. This would
  silently kill CI jobs.

  This is the likely explanation for the zombie processes you're seeing: if you ran the script from a
  wrapper or interactive shell, kill 0 partially killed the process tree but left some orphaned children
  whose parent died mid-execution.

  Fix: Replace kill 0 with something that only targets this script's descendants:

  trap "rm -fr $TMPDIR || true; kill -- -$$ 2>/dev/null || true" EXIT
  trap "exit 1" INT TERM

  kill -- -$$ sends SIGTERM to the process group whose PGID equals this script's PID — but only works if
  the script is the process group leader. A more portable approach:

  cleanup() {
    rm -fr "$TMPDIR" || true
    # Kill only our direct background children, not the parent's process group.
    jobs -p | xargs -r kill 2>/dev/null || true
    wait 2>/dev/null || true
  }
  trap cleanup EXIT
  trap "exit 1" INT TERM

  Splitting INT/TERM from EXIT avoids re-entrancy: INT/TERM just set the exit code, then EXIT runs once for actual cleanup.
```

.rhdh/scripts/install-rhdh-catalog-source.sh
```diff
@@ -556,8 +624,9 @@ fi
 TMPDIR=$(mktemp -d)
 pushd "${TMPDIR}" > /dev/null
 debugf ">>> WORKING DIR: $TMPDIR <<<"
+
 # shellcheck disable=SC2064
-trap "rm -fr $TMPDIR || true" EXIT
+trap "rm -fr $TMPDIR || true; kill 0" EXIT INT TERM
```
rm3l:
```suggestion
trap "rm -fr $TMPDIR || true; jobs -p | xargs -r kill 2>/dev/null; wait 2>/dev/null" EXIT
trap "exit 1" INT TERM
```

Suggestion from Claude, which seems to be working slightly better as I can still see some leftover processes (as before). Maybe we can apply this suggestion, and we can investigate those zombie processes later in a followup issue/PR.

subhashkhileri: Thanks for the suggestion, fixed it.

rm3l (approved):
Thanks @subhashkhileri !!

rm3l:
/override "PR Publish"
/override "PR Validate"

unrelated

rm3l:
/cherry-pick release-1.10

Just as a reminder to backport to 1.10 once 1.10.0 is out.

PR #2909, opened 2026-05-27
Make empty flavours: [] disable Default Flavours

gazarenkov:
## Description

This is the fix for the issue when setting spec.flavours: [] (empty array) in a Backstage CR does not disable default flavours as documented. 

https://redhat.atlassian.net/browse/RHDHBUGS-3198

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

1. Create a Backstage CR with explicit empty flavours array:

```
kind: Backstage
metadata:
  name: test-backstage
spec:
  flavours: []  # Explicitly disable all flavours
```
2. Check the deployed pod's configuration
3. Observe that default flavours (e.g., lightspeed) not loaded


config/profile/rhdh/default-config/deployment.yaml
```diff
@@ -70,7 +70,7 @@ spec:
               value: "40000000"
             # CATALOG_INDEX_IMAGE will be replaced by the value of the `RELATED_IMAGE_catalog_index` env var, if set
             - name: CATALOG_INDEX_IMAGE
-              value: "quay.io/rhdh/plugin-catalog-index:1.10"
+              value: "quay.io/rhdh/plugin-catalog-index:next"
```
Fortune-Ndlovu, line 73: I guess when we backport this fix to the release-1.0 branch, we need to use the 1.10 tag

pkg/model/flavour.go
```diff
@@ -39,15 +39,24 @@ func GetEnabledFlavours(spec api.BackstageSpec) ([]enabledFlavour, error) {
 		return nil, err
 	}
 
+	// Debug logging
+	if spec.Flavours == nil {
+		fmt.Println("DEBUG GetEnabledFlavours: spec.Flavours=nil (not specified, using defaults)")
```
Fortune-Ndlovu: should these `fmt.Println` / `fmt.Printf` use structured logger instead?

pkg/model/flavour.go
```diff
@@ -39,15 +39,24 @@ func GetEnabledFlavours(spec api.BackstageSpec) ([]enabledFlavour, error) {
...
+
 	// Step 2: Override enabled status from spec
 	if spec.Flavours != nil {
+		flavours := *spec.Flavours
 
-		// No specified returns empty list
-		if len(spec.Flavours) == 0 {
+		// Empty array explicitly disables all flavours
+		if len(flavours) == 0 {
+			fmt.Println("DEBUG: Empty flavours array specified - disabling all flavours")
```
Fortune-Ndlovu: same here this fmt.Println should it be converted to structured logging for consistency?

integration_tests/suite_test.go
```diff
@@ -172,11 +172,13 @@ func createBackstage(ctx context.Context, spec api.BackstageSpec, ns string, nam
...
 
 func createAndReconcileBackstage(ctx context.Context, ns string, spec api.BackstageSpec, name string) string {
+
 	backstageName, err := createBackstage(ctx, spec, ns, name)
 	Expect(err).To(Not(HaveOccurred()))
 
+	found := &api.Backstage{}
 	Eventually(func() error {
-		found := &api.Backstage{}
+		//found := &api.Backstage{}
```
Fortune-Ndlovu: Should this line be just deleted, we leaving the old declaration as a comment right below the moved one.

integration_tests/suite_test.go
```diff
@@ -196,6 +198,19 @@ func createAndReconcileBackstage(ctx context.Context, ns string, spec api.Backst
 	return backstageName
 }
 
+// TODO: it is not working as Backstage app is starting faster than DB and failed.
```
Fortune-Ndlovu, line 200: should we create a jira issue for this and add this jira link in this comment so we can track both ends

Fortune-Ndlovu (commented):
When testing locally, I received no errors, and reconciliation completed cleanly. Below are just nitpicks worth considering

config/profile/rhdh/default-config/deployment.yaml
gazarenkov: Yes, there is a certain point of release/post-release (create release-X.Y branch, fix versions etc) but meantime I think we should use "next" same as we do for rhdh image?

integration_tests/suite_test.go
gazarenkov: yes, will do, just want to do a short experiment before to make sure

pkg/model/flavour.go
gazarenkov: Same, removed

pkg/model/flavour.go
gazarenkov: Good catch, thanks! Just forgot to remove some debugging info. Fixed

config/profile/rhdh/default-config/deployment.yaml
Fortune-Ndlovu: yeah I agree. Also looking at the `release-1.10` branch https://github.com/redhat-developer/rhdh-operator/blob/release-1.10/config/profile/rhdh/default-config/deployment.yaml#L73, I think on `main` branch it makes sense to use `next`. Also sidenote do you think we should also have a `next-1.10` tag for the plugin-catalog-index similar to the `quay.io/rhdh-community/rhdh:next-1.10` image?

Fortune-Ndlovu (approved):
lgtm

PR #2918, opened 2026-05-28
 Refactor runtime object handling 

gazarenkov:
## Description

 This PR refactors the core operator runtime model to simplify object initialization flow

  Key Changes

  1. Simplified RuntimeObject Interface
  - Consolidated addToModel() signature: now takes config parameter directly instead of separate setObject() call
  - Merged setMetaInfo() into updateAndValidate() phase
  - Added GetKey() method for object type identification
  - Removed ExternalConfigContributor interface - functionality integrated into main flow
  - Removed the intermediate setObject() and separate setMetaInfo() steps.

  2. Simplified Initialization Flow (3 phases → 2 phases)
  - Phase 1: Choose config (overlay OR default, not both), initialize object via addToModel(), register in model
  - Phase 2: Update references, validate, set metadata via updateAndValidate()

  3. Deterministic Object Ordering
  - Objects maintain registration order 

  4. Cleaner API
  - Removed key parameter from setRuntimeObject(key, obj) → setRuntimeObject(obj)
  - Objects self-identify via GetKey() method
  - GetRuntimeObjects() returns slice instead of map for predictable iteration

 
https://redhat.atlassian.net/browse/RHDHBUGS-2975

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

It does not change the behaviour


pkg/model/runtime.go
```diff
@@ -129,74 +146,62 @@ func InitObjects(ctx context.Context, backstage api.Backstage, externalConfig Ex
...
-		// read configuration defined in BackstageCR.Spec.RawConfigContent ConfigMap
-		// if present, backstageObject's default configuration will be overridden
+		// First, try overlay from CR spec
 		overlay, overlayExist := externalConfig.RawConfig[conf.Key]
 		if overlayExist {
-			// new object to replace default, not merge
 			if objs, err := utils.ReadYamls([]byte(overlay), nil, *scheme); err != nil {
 				if !errors.Is(err, os.ErrNotExist) {
-					return nil, fmt.Errorf("failed to read overlay value for the key %s, reason: %w", conf.Key, err)
+					return nil, fmt.Errorf("failed to read overlay config for the key %s, reason: %s", conf.Key, err)
```
rm3l:
```suggestion
					return nil, fmt.Errorf("failed to read overlay config for the key %s, reason: %w", conf.Key, err)
```

Might be helpful for debugging, as done in https://github.com/redhat-developer/rhdh-operator/pull/2803

pkg/model/runtime.go
```diff
@@ -129,74 +146,62 @@ func InitObjects(ctx context.Context, backstage api.Backstage, externalConfig Ex
...
-			backstageObject.setMetaInfo(backstage, scheme)
-		}
-		if ecc, ok := backstageObject.(ExternalConfigContributor); ok {
-			// backstageObject is an instance of ExternalConfigContributor
-			ecs = append(ecs, ecc)
+		// If no overlay, use default config
+		if chosenConfig == nil {
+			if objs, err := ReadDefaultConfig(conf, flavours, *scheme, platform.Extension); err != nil {
+				if !errors.Is(err, os.ErrNotExist) {
+					return nil, fmt.Errorf("failed to read default value for the key %s, reason: %s", conf.Key, err)
```
rm3l:
```suggestion
					return nil, fmt.Errorf("failed to read default value for the key %s, reason: %w", conf.Key, err)
```

pkg/model/runtime.go
```diff
@@ -129,74 +146,62 @@ func InitObjects(ctx context.Context, backstage api.Backstage, externalConfig Ex
...
-	}
-
-	// set generic metainfo and updateAndValidate all
-	for _, v := range model.RuntimeObjects {
-		err := v.updateAndValidate(backstage)
-		if err != nil {
-			return nil, fmt.Errorf("failed object validation, reason: %w", err)
+		// Add object to model (always added, even if config is nil - placeholder pattern)
+		if err := backstageObject.addToModel(model, backstage, chosenConfig, scheme); err != nil {
+			return nil, fmt.Errorf("failed to add object to model for key %s, reason: %s", conf.Key, err)
```
rm3l:
```suggestion
			return nil, fmt.Errorf("failed to add object to model for key %s, reason: %w", conf.Key, err)
```

pkg/model/runtime.go
```diff
@@ -129,74 +146,62 @@ func InitObjects(ctx context.Context, backstage api.Backstage, externalConfig Ex
...
-	for _, ecc := range ecs {
-		if err := ecc.addExternalConfig(backstage.Spec); err != nil {
-			return nil, fmt.Errorf("failed to contribute external config, reason: %w", err)
+	// Phase 2: updateAndValidate all objects
+	// All objects are now in model, so cross-references are safe
+	// Iterate over RuntimeObjects in their registration order (deterministic)
+	for _, obj := range model.RuntimeObjects {
+		err := obj.updateAndValidate(backstage, scheme)
+		if err != nil {
+			return nil, fmt.Errorf("failed object validation, reason: %s", err)
```
rm3l:
```suggestion
			return nil, fmt.Errorf("failed object validation, reason: %w", err)
```

pkg/model/dynamic-plugins.go
```diff
@@ -37,6 +37,10 @@ func (f DynamicPluginsFactory) newBackstageObject() RuntimeObject {
 type DynamicPlugins struct {
 	ConfigMap *corev1.ConfigMap
 	model     *BackstageModel
+	//EffectivePlugins      *corev1.ConfigMap
+	//EffectiveAppConfig    *corev1.ConfigMap
+	//EffectiveDependencies []PluginDependency
+	//dynaPlugins           []DynaPlugin
```
rm3l, line 43:
```suggestion
```

I guess these can be removed?

pkg/model/runtime.go
```diff
@@ -41,57 +56,54 @@ type BackstageModel struct {
...
-	localDbStatefulSet *DbStatefulSet
-	LocalDbService     *DbService
-	LocalDbSecret      *DbSecret
-
-	route          *BackstageRoute
-	appConfig      *AppConfig
-	DynamicPlugins DynamicPlugins
-
+	// RuntimeObjects contains all object wrappers in deterministic order.
+	// Order is determined by init() registration (alphabetical by source file name).
```
rm3l: > alphabetical by source file name

I hope we are not relying on this for correctness, otherwise it might cause issues if someone were to rename any of the source files in the future :-)
But I didn't see anything actually depending on this specific alphabetical ordering, so all good.

pkg/model/runtime.go
rm3l, line 114: With this refactoring leaning on keys, it might make sense to check here at registration time if a key already exists and panic, instead of just appending blindly?
Otherwise, if someone accidentally registered the same key twice (say, two `init()` functions using `DeploymentKey`), you'd get 2 entries in `runtimeConfig`, 2 wrappers in `RuntimeObjects`, and `GetRuntimeObject` would silently return whichever was added first (depending on the alphabetical source file name).
I guess a panic is appropriate here since `registerConfig` is only called from `init()`, so it would fail right away in that case at startup.
WDYT?

rm3l (commented):
Looks good overall. A few comments below.

rm3l:
Also, some of the review comments from Qodo seem relevant to address, at least the `Action required` and `Review recommended` points: https://github.com/redhat-developer/rhdh-operator/pull/2918#issuecomment-4594009957

pkg/model/dynamic-plugins.go
gazarenkov: I intentionaly kept it as supposed to be used in the next PR

gazarenkov:
> Also, some of the review comments from Qodo seem relevant to address, at least the `Action required` and `Review recommended` points: [#2918 (comment)](https://github.com/redhat-developer/rhdh-operator/pull/2918#issuecomment-4594009957)

I think I fixed both Action required? Can't you see it?

pkg/model/runtime.go
gazarenkov: Well, it is not possible runtime :) only if developer intentionally change the object set with mistake.
But, sure, let's improve it even for this case, thanks

pkg/model/runtime.go
gazarenkov: remove this comment to avoid confusing for the time

rm3l (approved):

PR #2727, opened 2026-04-26
restore v1alpha1, v1alpha2 and move v1alpha6 to v1alpha5

gazarenkov:
## Description

- restore versions v1alpha1, v1alpha2 
- make versions v1alpha1, v1alpha2, v1alpha3 unserved and deprecated
- move new fields (Flavours) to  v1alpha5 and make it (back) storage version
- remove v1alpha6 

## Which issue(s) does this PR fix or relate to

https://redhat.atlassian.net/browse/RHDHBUGS-2993

## PR acceptance criteria

- [x] Tests
- [x] Documentation

PR #2998, opened 2026-06-12
fix(ci): use dynamic API version in upgrade test

rm3l:
## Description

The upgrade test hardcoded `rhdh.redhat.com/v1alpha3` when creating the Backstage CR against the "from" operator. Since 1.10+ no longer serves `v1alpha3` (https://github.com/redhat-developer/rhdh-operator/pull/2727), the `1.10 => main` upgrade path fails: https://github.com/redhat-developer/rhdh-operator/actions/runs/27315253434/job/80694281540

This PR queries the installed CRD for its storage version instead, which is always served regardless of the branch.

Will need to be cherry-picked to `release-1.10` as well.

- Fixes https://redhat.atlassian.net/browse/RHDHBUGS-3349

## PR acceptance criteria

- [x] Tests
- [ ] Documentation


rm3l:
Merging so we can manually trigger the upgrade nightly checks..

PR #3091, opened 2026-06-25
chore: remove EOL 1.8 from the nightly checks and disable in Renovate

rm3l:
## Description
1.8 is EOL with the release of 1.10.0. But note that we are still keeping it in the upgrade checks because we should be testing upgrades from an unsupported version as well to make sure it works as expected

- Fixes https://redhat.atlassian.net/browse/RHIDP-15110

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

Will check after merging.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -26,6 +26,12 @@
   "packageRules": [
     {
       "matchBaseBranches": [
+        "release-1.8"
+      ],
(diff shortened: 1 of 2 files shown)


rm3l:
Merging to trigger the nightly checks..

PR #3234, opened 2026-07-21
chore: use centralised TLS Profile

OpinionatedHeron:
## Description
Implemented centralized OpenShift TLS security profile support for the RHDH operator (controller-runtime path via github.com/openshift/controller-runtime-common), so the operator’s TLS servers honor apiservers.config.openshift.io/cluster. 

- Fixes [RHIDP-13852](https://redhat.atlassian.net/browse/RHIDP-13852)

## PR acceptance criteria

- [x] Tests
- [x] Documentation


cmd/main.go
```diff
@@ -253,6 +290,30 @@ func main() {
 	}
 	// +kubebuilder:scaffold:builder
 
+	// Watch for TLS profile changes and restart so the new config is applied.
+	if plf.IsOpenshift() {
+		if err := (&tlspkg.SecurityProfileWatcher{
```
gazarenkov: Same, could you please consider to pick this block to standalone function?

cmd/main.go
```diff
@@ -91,10 +96,42 @@ func main() {
...
+	ctx, cancel := context.WithCancel(ctrl.SetupSignalHandler())
+	defer cancel()
+
 	if metricsAddr != "0" && !secureMetrics {
 		setupLog.Info("Metrics are served over plaintext HTTP. This is only intended for local development.")
 	}
 
+	restConfig := ctrl.GetConfigOrDie()
+
+	// Fetch the TLS profile from apiservers.config.openshift.io/cluster.
```
gazarenkov: Can we refactor a bit this block to reduce the main() and avoid unnecessary calls and log messages if not Openshift.
For ex: 
- call `plf, err := controller.DetectPlatform()` earlier, check it and call tlspkg related code only if Openshift
- instead of 30 lines (or so) inline make a function to return tlsSecurityProfileSpec, tlsAdherence, tlsConfig

OpinionatedHeron: Thanks so much, Gennady. I have updated the file now, and moved the code block to it's own function. This should reduce main.

gazarenkov (approved):

PR #3289, opened 2026-07-27
plugin installer

gazarenkov:
## Description

Added plugin-installer/ - a container image for downloading dynamic plugins with parallel execution support.

  Image: quay.io/rhdh-community/plugin-installer:next

  Script features:
  - Downloads from OCI registries, HTTP/HTTPS, and NPM
  - Parallel downloads (default: 4 concurrent)
  - Skopeo for OCI (FIPS compliant), oras as fallback
  - NPM .npmrc parsing for private registries
  - SHA integrity verification (sha256/sha384/sha512)
  - Lock file for concurrent safety
  - SIGTERM forwarding to child processes

  Makefile targets: install-dp-build, install-dp-push

  CI: Tests on PR, builds/pushes on merge to main

https://redhat.atlassian.net/browse/RHIDP-15359

## PR acceptance criteria

- [x] Tests
- [x] Documentation

Run it with 'make install run' (cluster deployment is not working yet)
Apply your CR


config/profile/rhdh/patches/deployment-patch.yaml
```diff
@@ -28,6 +28,8 @@ spec:
               value: quay.io/fedora/postgresql-15:latest
             - name: RELATED_IMAGE_backstage
               value: quay.io/rhdh-community/rhdh:next
+            - name: INSTALL_DP_IMAGE
```
rm3l, line 31: Can this follow the same `RELATED_IMAGE_` prefix pattern, so that it can be reflected in the CSV's `spec.relatedImages` and handled automatically by tools like `oc-mirror`?

plugin-installer/install_plugins.sh
```diff
@@ -0,0 +1,714 @@
...
+download_http() {
+    local url="$1"
+    local plugin_name="$2"
+    local plugin_dir="$3"
+    local integrity="${4:-}"
+
+    local tmp_file
+    tmp_file=$(mktemp)
+
+    if ! curl -sL "${url}" -o "${tmp_file}"; then
```
rm3l:
```suggestion
    if ! curl --fail -sL "${url}" -o "${tmp_file}"; then
```

Otherwise, I think it might return 0 even on HTTP 404/500 errors.

plugin-installer/install_plugins.sh
```diff
@@ -0,0 +1,714 @@
...
+    if [[ -n "${NPM_AUTH_TOKEN}" ]]; then
+        auth_header="Authorization: Bearer ${NPM_AUTH_TOKEN}"
+    fi
+
+    # Fetch package metadata
+    local metadata
+    metadata=$(mktemp)
+
+    if [[ -n "${auth_header}" ]]; then
+        if ! curl -sL -H "${auth_header}" "${registry_url}" -o "${metadata}"; then
```
rm3l:
```suggestion
        if ! curl --fail -sL -H "${auth_header}" "${registry_url}" -o "${metadata}"; then
```

plugin-installer/install_plugins.sh
```diff
@@ -0,0 +1,714 @@
...
+            ;;
+    esac
+
+    # Find layer files (skip manifests/config)
+    # shellcheck disable=SC2012 # OCI blob filenames are SHA256 hashes (alphanumeric only)
+    local layer_file
+    if [[ "${OCI_TOOL}" == "oras" ]]; then
+        layer_file=$(find "${tmp_oci}/blobs/sha256" -type f -print0 | xargs -0 ls -S 2>/dev/null | head -1)
+    else
+        layer_file=$(find "${tmp_oci}" -maxdepth 1 -type f ! -name "*manifest*" ! -name "*version*" -print0 | xargs -0 ls -S 2>/dev/null | head -1)
```
rm3l, line 237: IIUC, this uses the largest layer, right? While it will work for single-layer images, isn't it possible that it silently picks the wrong blob? Maybe verify the media type too from the manifest instead, or at least verify the tar contains the expected content structure before proceding?

plugin-installer/install_plugins.sh
```diff
@@ -0,0 +1,714 @@
...
+    metadata=$(mktemp)
+
+    if [[ -n "${auth_header}" ]]; then
+        if ! curl -sL -H "${auth_header}" "${registry_url}" -o "${metadata}"; then
+            echo "[FAIL] ${plugin_name}: failed to fetch package metadata" >&2
+            rm -f "${metadata}"
+            return 1
+        fi
+    else
+        if ! curl -sL "${registry_url}" -o "${metadata}"; then
```
rm3l:
```suggestion
        if ! curl --fail -sL "${registry_url}" -o "${metadata}"; then
```

plugin-installer/install_plugins.sh
```diff
@@ -0,0 +1,714 @@
...
+
+    if [[ "${computed}" != "${hash}" ]]; then
+        echo "Integrity check failed: expected ${hash}, got ${computed}" >&2
+        return 1
+    fi
+
+    return 0
+}
+
+download_npm() {
```
rm3l, line 397: From Claude: this fetches the full package document but parses it with flat grep — risks wrong-field matches.

The registry URL `${NPM_REGISTRY}/${encoded_name}` returns the complete package document (every version ever published, readme text, keywords, etc.). The script then greps through that entire blob for patterns like `"tarball":"...<version>..." and "latest":"..."`.

These patterns aren't anchored to any JSON structure, so they can match inside unrelated fields. The tarball grep is the most concerning: version 1.0 could substring-match a tarball URL for 1.0.1, silently downloading the wrong version.

NPM's registry exposes a version-specific endpoint that returns only the metadata for one version — small, flat, with exactly one tarball and one integrity:  `GET ${registry}/${package}/${version}`

**Suggestion**: resolve `latest` from the full doc (unavoidable), then fetch the version-specific endpoint for tarball URL and integrity. That eliminates the ambiguity without adding any dependencies.

.github/workflows/plugin-installer.yaml
rm3l, line 1: This should also be triggered on the release branches, IMO.

plugin-installer/install_plugins.sh
gazarenkov: Fixed it this way:
 - npm with explicit version (is-odd@3.0.1) - skips full doc, fetches version-specific endpoint directly
  - npm with latest resolution (is-even) - fetches full doc → resolves "latest" → fetches version-specific endpoint

plugin-installer/install_plugins.sh
gazarenkov: Added validation on io.backstage.dynamic-packages annotation - so only artifacts created by rhdh-cli accepted

config/profile/rhdh/patches/deployment-patch.yaml
gazarenkov: Good point, let me think about it more.
Actually it does not work for the time on prod (requires Catalog CR feature), just placeholder.
Thanks

.github/workflows/plugin-installer.yaml
gazarenkov: added

.github/workflows/plugin-installer.yaml
```diff
@@ -0,0 +1,74 @@
+name: Plugin Installer CI
+
+on:
+  push:
+    branches:
+      - main
+      - 'release-[2-9].[0-9]*'
```
rm3l:
```suggestion
      - 'release-[2-9]+.[0-9]+'
```

not problematic, but just to be consistent with the other workflows: https://github.com/redhat-developer/rhdh-operator/blob/main/.github/workflows/next-container-build.yaml#L8

.github/workflows/plugin-installer.yaml
```diff
@@ -0,0 +1,74 @@
...
+    paths:
+      - 'plugin-installer/**'
+      - '.github/workflows/plugin-installer.yaml'
+  pull_request:
+    branches:
+      - main
```
rm3l, line 13:
```suggestion
      - main
      - 'release-[2-9]+.[0-9]+'
```

rm3l (approved):

rm3l (approved):

PR #3290, opened 2026-07-29
chore: integrate crdify CRD upgrade safety checks into CI [RHIDP-8670]

Fortune-Ndlovu:
## Description

OLM v1[mod.go file](https://github.com/operator-framework/operator-controller/blob/main/go.mod#L53) uses crdify (https://github.com/kubernetes-sigs/crdify)  to perform CRD upgrade safety preflight checks at runtime. If the RHDH operator's CRD has breaking changes between versions, OLM v1 rejects the upgrade on customer clusters see https://github.com/operator-framework/operator-controller/blob/main/internal/operator-controller/rukpak/preflights/crdupgradesafety/crdupgradesafety.go. This PR integrates crdify into CI so we catch these issues at PR time rather than at runtime.

- Fixes https://redhat.atlassian.net/browse/RHIDP-8670 

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

```bash
# 1. Check against main (default baseline)
make crd-upgrade-check

# 2. Check against a specific release branch
make crd-upgrade-check CRD_BASELINE_REF=release-1.10

# 3. Negative test temporarily break the CRD, verify crdify catches it
sed -i '/^\s*replicas:/d' config/crd/bases/rhdh.redhat.com_backstages.yaml
make crd-upgrade-check
# Should FAIL with NoExistingFieldRemoved

# 4. Restore the CRD after negative test
git checkout config/crd/bases/rhdh.redhat.com_backstages.yaml
```


.github/workflows/pr-crd-upgrade-check.yaml
```diff
@@ -0,0 +1,42 @@
...
+    timeout-minutes: 10
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
```
rm3l:
```suggestion
```

Would suggest removing this, so that the checkout step runs by default against the merge commit, which is actually better in case `main` moved forward and the base CRD changed since the PR was opened.
Also, `fetch-depth: 0` would fetch the entire repo history, which seems unnecessary here IMO since the base branch (which is all we need regardless of fetch depth) is being fetched below.

Makefile
```diff
@@ -226,6 +226,14 @@ validate-image-digests: ## Validate that Dockerfile digests are manifest lists (
 	@echo
 	@hack/validate-image-digests.sh
 
+CRD_FILE ?= config/crd/bases/rhdh.redhat.com_backstages.yaml
+CRD_BASELINE_REF ?= main
+
+.PHONY: crd-upgrade-check
+crd-upgrade-check: crdify ## Check CRD upgrade safety against the base branch.
+	@echo "Checking CRD upgrade safety: $(CRD_BASELINE_REF) -> current working tree"
+	@$(CRDIFY) "git://$(CRD_BASELINE_REF)?path=$(CRD_FILE)" "file://$(CRD_FILE)"
```
rm3l: So this runs against `config/crd/bases/rhdh.redhat.com_backstages.yaml`. But actually, the CRD manifest is the output of a kustomization project in `config/crd/kustomization.yaml`, which may apply additional patches that could alter the final CRD.
I would suggest running `crdify` against the output of `$(KUSTOMIZE) build config/crd/` instead. This might require cloning the base branch and generating the effective CRD to a temporary file before comparing, but I think it would provide a cleaner and accurate verification.

.github/workflows/pr-crd-upgrade-check.yaml
Fortune-Ndlovu: Good catch, removed the explicit `ref` and `fetch-depth: 0` so checkout uses the default merge commit, and we still fetch only the base branch for the baseline CRD.

Makefile
Fortune-Ndlovu: Agreed, switched the check to compare `$(KUSTOMIZE) build config/crd/` for both the baseline (via `git archive`) and the current tree, so any CRD patches are included. This also matches how we already install the CRDs.

Makefile
```diff
@@ -227,6 +227,19 @@ validate-image-digests: ## Validate that Dockerfile digests are manifest lists (
 	@echo
 	@hack/validate-image-digests.sh
 
+CRD_BASELINE_REF ?= main
+
+.PHONY: crd-upgrade-check
+crd-upgrade-check: crdify kustomize ## Check CRD upgrade safety against the base branch.
+	@echo "Checking CRD upgrade safety: $(CRD_BASELINE_REF) -> current working tree"
+	@tmpdir=$$(mktemp -d); \
```
rm3l, line 235:
```suggestion
	@tmpdir=$$(mktemp -d); \
	set -e; \
```

To avoid silently passing if any of the previous commands fail, which could result in a potentially false pass, I guess.

.github/workflows/pr-crd-upgrade-check.yaml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches:
+      - 'main'
+      - 'release-[0-9]*.[0-9]*'
+    paths:
+      - '.github/workflows/pr-crd-upgrade-check.yaml'
+      - 'config/crd/**'
+      - 'api/**'
```
rm3l, line 12:
```suggestion
      - 'api/**'
      - 'Makefile'
```

Should also include the Makefile in case the `crd-upgrade-check` or `crdify` targets are changed, or in case the version of crdify is updated.

.github/workflows/pr-crd-upgrade-check.yaml
```diff
@@ -0,0 +1,39 @@
+name: PR CRD Upgrade Safety Check
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches:
+      - 'main'
+      - 'release-[0-9]*.[0-9]*'
```
rm3l:
```suggestion
      - 'release-[0-9]+.[0-9]+'
```

Too broad IMO. This should be the right pattern., similar to https://github.com/redhat-developer/rhdh-operator/blob/main/.github/workflows/pr-bundle-diff-checks.yaml#L8

rm3l (approved):

PR #3318, opened 2026-08-04
feat: add `rhdh.redhat.com/idle` annotation support for idling workloads [RHIDP-15995]

rm3l:
## Description

Adds support for the `rhdh.redhat.com/idle` annotation on the Backstage CR. When set to `"true"`, the operator sets replicas to 0 on the Backstage Deployment/StatefulSet and (if enabled) the local DB StatefulSet. Removing the annotation lets the normal reconcile loop rebuild the model from scratch, which restores the default replica count (or the user's patched value).

This enables external controllers (e.g. Dev Sandbox idler) to scale RHDH instances to zero without conflicting with the operator's Server-Side Apply field ownership of the replicas field.

**External DB compatibility:** When `spec.database.enableLocalDb=false`, only the Backstage Deployment is affected. The external DB is never touched.

## Which issue(s)

Fixes [RHIDP-15995](https://issues.redhat.com/browse/RHIDP-15995)

Part of: [RHIDP-15996](https://issues.redhat.com/browse/RHIDP-15996), [RHDHPLAN-1040](https://issues.redhat.com/browse/RHDHPLAN-1040)

## PR acceptance criteria

- [x] Tests
- [x] Documentation

```bash
# Tests should pass
make test

# Manual test on a cluster
oc annotate backstage developer-hub rhdh.redhat.com/idle=true
# Verify pods scale to 0, status condition shows reason=Idled
oc annotate backstage developer-hub rhdh.redhat.com/idle-
# Verify pods scale back (replicas restored from default config or spec.deployment.patch)
```


gazarenkov (changes requested):
Why that complex?
IMO the logic is just:
- check the annotation (in corresponding addToModel or updateAndValidate methods)
- if idled set replica=0 in deployment, statefulset
- otherwise keep unchanged
Did I miss something?

rm3l:
> Why that complex? IMO the logic is just:
> 
>     * check the annotation (in corresponding addToModel or updateAndValidate methods)
> 
>     * if idled set replica=0 in deployment, statefulset
> 
>     * otherwise keep unchanged
>       Did I miss something?

Yeah, good point. This is just the initial draft Claude implementation for this. Still refining it :). Thanks.

docs/configuration.md
```diff
@@ -913,6 +914,59 @@ spec:
             $patch: delete
 ```
 
+### Instance Idling
```
gazarenkov: I am not sure configuration.yaml is the best place for it. 
Let's better consider admin.md ?

docs/configuration.md
```diff
@@ -913,6 +914,59 @@ spec:
...
+
+The Operator supports idling and waking Backstage instances via the `rhdh.redhat.com/idle` annotation on the Backstage CR. This is useful for environments like Dev Sandbox where instances should be scaled to zero when inactive.
+
+#### How it works
+
+When the annotation `rhdh.redhat.com/idle` is set to `"true"` on the Backstage CR, the Operator overrides replicas to 0 on both the Backstage Deployment (or StatefulSet) and the local DB StatefulSet (if enabled), in the same namespace as the Backstage CR. The status condition reason is set to `Idled`.
+
+When the annotation is removed (or set to any value other than `"true"`), the Operator restores replicas:
+- If the user specified replicas via `spec.deployment.patch`, that value is preserved.
+- Otherwise, replicas defaults to 1.
```
gazarenkov: There should not be any diff specified or not, Backstage spec should be untouchable.
Let's consider removing this section, it is confusing.

integration_tests/idle_test.go
```diff
@@ -0,0 +1,171 @@
...
+		}, time.Minute, time.Second).Should(Succeed())
+
+		By("verifying status condition is no longer Idled")
+		Eventually(func(g Gomega) {
+			g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: backstageName, Namespace: ns}, bs)).To(Succeed())
+			g.Expect(bs.Status.Conditions[0].Reason).NotTo(Equal("Idled"))
+		}, time.Minute, time.Second).Should(Succeed())
+	})
+
+	It("wakes with user-specified replicas from deployment patch", func() {
```
gazarenkov: Do we really need this test? 
It is a bit redundant IMO?

pkg/model/deployment_obj.go
```diff
@@ -56,6 +56,10 @@ func (d *DeploymentObj) SpecReplicas() *int32 {
 	return d.Obj.Spec.Replicas
 }
 
+func (d *DeploymentObj) SetReplicas(r *int32) {
```
gazarenkov: May be better to have something like Idle() function instead?
IMO it would be more straightforward and we hardly need replicas to be changed internally by other reason?

docs/configuration.md
rm3l: makes sense. Let me move it there.

docs/configuration.md
rm3l: Good catch. That was a leftover from the previous draft implementation. Will update.

integration_tests/idle_test.go
rm3l: removed. leftover from the previous draft implementation.

pkg/model/runtime.go
```diff
@@ -205,6 +206,15 @@ func InitObjects(ctx context.Context, backstage api.Backstage, externalConfig Ex
 		}
 	}
 
+	// Phase 3: idle all workloads if the annotation requests it
```
gazarenkov: Why overcomplicate it with additional abstraction?
Idle annotation is a part of CR config and known in advance (not calculated above in InitObjects()).
I think you can just call something like:
if backstage.GetAnnotations()[IdleAnnotation] == "true" {
  idle() - for db ss or deployable.idle() for deployment
}
at the end of addToModel() (after setDeployment in deployment - it is the only constraint I guess)?

docs/configuration.md
```diff
@@ -29,6 +29,7 @@ It is highly recommended to read the [Design](design.md) document to understand
   - [Deployment Configuration](#deployment-configuration)
     - [Deployment Kind](#deployment-kind) 
     - [Deployment Patching](#deployment-patching)
+  - [Instance Idling](#instance-idling)
```
gazarenkov: Not a valid link, right?

pkg/model/runtime.go
rm3l: okay, let's simplify. I initially thought we could be more generic to prepare for future "idle-able" resources. But sure, let's start simple and generalize later if need be.

rm3l: Simplified in [`947d880` (this PR)](https://github.com/redhat-developer/rhdh-operator/pull/3318/commits/947d88070e9ee4613381c5fe474a392b9fbadbbb). Thanks.

internal/controller/backstage_controller.go
```diff
@@ -115,6 +115,11 @@ func (r *BackstageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, errorAndStatus(&backstage, "failed to apply backstage objects", err)
 	}
 
+	if backstage.GetAnnotations()[model.IdleAnnotation] == "true" {
+		setStatusCondition(&backstage, api.BackstageConditionTypeDeployed, metav1.ConditionFalse, api.BackstageConditionReasonIdled, "Instance is idled")
+		return ctrl.Result{}, nil
```
gazarenkov: I think it should not prevent checking deployment status below, otherwise it is just duplication of what is declared in annotation, not the current deployment state.
Consider setting the reason/message of Notdeployed inside setDeploymentStatus instead.

rm3l: Makes sense, thanks. Updated in [`05406da` (this PR)](https://github.com/redhat-developer/rhdh-operator/pull/3318/commits/05406daf397c02e05e9302d05969a39fc5a6eb97).

gazarenkov (approved):

PR #3348, opened 2026-08-07
CR Status Improvements

gazarenkov:
## Description

- Two-level health monitoring: Added Runtime condition alongside existing Deployed condition
    Deployed: Reflects Deployment/StatefulSet replica status
    Runtime: Provides Pod/Container-level health (crash loops, image pull failures, OOMKilled, etc.)
 - Improved status messages
  - Smart crash loop detection: Only reports crash if container is NOT currently Ready - containers that recover after restart are not flagged as failed
  - Enabled plugins in status: status.plugins field lists enabled dynamic plugins when healthy, cleared when unhealthy
  - Unit tests: Added ~490 lines of tests covering all container state scenarios
  - Updated docs/design.md with full status condition reference tables and examples

https://redhat.atlassian.net/browse/RHIDP-15021

## PR acceptance criteria

- [x] Tests
- [x] Documentation


internal/controller/backstage_controller.go
```diff
@@ -90,37 +91,41 @@ func (r *BackstageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	if err != nil {
-		return ctrl.Result{}, errorAndStatus(&backstage, "failed to apply backstage objects", err)
+		return ctrl.Result{}, errorAndStatus(&backstage, api.BackstageConditionTypeDeployed, api.BackstageConditionReasonFailed, "failed to apply backstage objects", err)
 	}
 
-	r.setDeploymentStatus(ctx, &backstage, *bsModel)
+	isReady := r.reconcileStatus(ctx, &backstage, *bsModel)
+	if !isReady {
+		// Requeue to check pod status again (for init container failures, etc.)
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
Fortune-Ndlovu, line 125: **Bug: Idled instances trigger infinite 10s requeue loop**

When a Backstage CR is idled (`rhdh.redhat.com/idle=true`), the deployment scales to 0 replicas:
- `setDeployedCondition` returns `true` (0/0 replicas ready = `Deployed`)
- `setRuntimeCondition` finds zero pods and returns `false` (`Pending: no pods found`)
- `isReady = true && false = false`
- Controller requeues every 10 seconds — indefinitely, for every idled instance

This generates unnecessary API server load and log noise.

Suggest checking for idle in `reconcileStatus` before evaluating runtime, or treating "no pods found" as healthy when the idle annotation is set.

internal/controller/backstage_controller.go
```diff
@@ -90,37 +91,41 @@ func (r *BackstageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// 2. Make some validation to fail fast
 	externalConfig, err := r.preprocessSpec(ctx, backstage)
 	if err != nil {
-		return ctrl.Result{}, errorAndStatus(&backstage, "failed to preprocess backstage spec", err)
+		return ctrl.Result{}, errorAndStatus(&backstage, api.BackstageConditionTypeConfig, api.BackstageConditionReasonInvalid, "failed to preprocess backstage spec", err)
```
Fortune-Ndlovu, line 94: **Config condition is set on error but never cleared on success**

When `preprocessSpec` or `InitObjects` fails, the `Config` condition is set to `Invalid` here. On the next successful reconciliation, this condition is never updated — a stale `Config=False/Invalid` will persist in the CR status even after the user fixes their config.

Either:
- Set `Config=True/Valid` on success (after line 105), or
- Remove the condition on success with `meta.RemoveStatusCondition`

api/v1alpha5/backstage_types.go
```diff
@@ -11,12 +11,23 @@ type BackstageConditionReason string
...
 const (
+	// Condition types
 	BackstageConditionTypeDeployed BackstageConditionType = "Deployed"
+	BackstageConditionTypeRuntime  BackstageConditionType = "Runtime"
+	BackstageConditionTypeConfig   BackstageConditionType = "Config"
 
+	// Deployed condition reasons
 	BackstageConditionReasonDeployed   BackstageConditionReason = "Deployed"
 	BackstageConditionReasonFailed     BackstageConditionReason = "DeployFailed"
 	BackstageConditionReasonInProgress BackstageConditionReason = "DeployInProgress"
```
Fortune-Ndlovu: **Breaking change: `BackstageConditionReasonIdled` removed**

`BackstageConditionReasonIdled` was a public API constant. Removing it breaks any external tooling or monitoring that matched on `reason: Idled`. The idle state is now encoded as `"(Idled)"` appended to the message string, which is harder to match programmatically.

Worth either:
- Keeping `Idled` as a distinct reason on the `Deployed` condition, or
- Documenting this as a breaking change in the PR description and release notes

internal/controller/backstage_status.go
```diff
@@ -33,22 +59,148 @@ func (r *BackstageReconciler) setDeploymentStatus(ctx context.Context, backstage
...
+	if isHealthy {
+		status = metav1.ConditionTrue
+	}
+	setStatusCondition(backstage, api.BackstageConditionTypeRuntime, status, reason, msg)
+
+	return isHealthy
+}
+
+// checkPodStates checks pods for errors, returns Running if all healthy.
+func (r *BackstageReconciler) checkPodStates(pods []corev1.Pod) (api.BackstageConditionReason, string) {
```
Fortune-Ndlovu: **Nit:** `checkPodStates` doesn't use the receiver `r` — should be a standalone function like `checkContainerState` below it. Makes it easier to unit test (no need to mock a reconciler) and is consistent with the rest of the file.

internal/controller/backstage_status_test.go
```diff
@@ -0,0 +1,491 @@
+package controller
```
Fortune-Ndlovu, line 1: Great coverage of `checkContainerState`, `deploymentState`, `statefulSetState`, and `setStatusCondition`.

Missing tests for `reconcileStatus`, `setRuntimeCondition`, and `checkPodStates` — these contain the pod-listing, multi-pod sorting, and overall readiness logic including the idle interaction (see the idle requeue bug in the controller). The codecov report shows 72 lines uncovered. Would be good to add table-driven tests for `checkPodStates` at minimum since it's a standalone-able function.

Fortune-Ndlovu (changes requested):
Great CR Status improvements. See inline comments for details and suggested fixes.

internal/controller/backstage_controller.go
gazarenkov: fixed

internal/controller/backstage_controller.go
gazarenkov: fixed

api/v1alpha5/backstage_types.go
gazarenkov: It is OK as it has never released.
"Idled" is not a deployment status, the status is calculated runtime and it is 0 replicas (pending), programmatically you can check/change the annotation

internal/controller/backstage_status.go
gazarenkov: fixed

internal/controller/backstage_status_test.go
gazarenkov: fixed

internal/controller/backstage_controller.go
```diff
@@ -90,37 +91,44 @@ func (r *BackstageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		// Requeue to check pod status again (for init container failures, etc.)
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+	}
 	return ctrl.Result{}, nil
 }
 
-func errorAndStatus(backstage *api.Backstage, msg string, err error) error {
-	setStatusCondition(backstage, api.BackstageConditionTypeDeployed, metav1.ConditionFalse, api.BackstageConditionReasonFailed, fmt.Sprintf("%s %s", msg, err))
+func errorAndStatus(backstage *api.Backstage, condType api.BackstageConditionType, reason api.BackstageConditionReason, msg string, err error) error {
+	setStatusCondition(backstage, condType, metav1.ConditionFalse, reason, err.Error())
```
rm3l:
```suggestion
	setStatusCondition(backstage, condType, metav1.ConditionFalse, reason, fmt.Sprintf("%s: %s", msg, err))
```

Might still be helpful to keep the descriptive message for more context, no?

internal/controller/backstage_status.go
```diff
@@ -14,41 +16,187 @@ import (
...
 	}
-	setStatusCondition(backstage, api.BackstageConditionTypeDeployed, status, state, msg)
+	setStatusCondition(backstage, api.BackstageConditionTypeRuntime, status, reason, msg)
+
+	return isHealthy
+}
+
+// checkPodStates checks pods for errors, returns Running/True if all healthy.
+func checkPodStates(pods []corev1.Pod) (metav1.ConditionStatus, api.BackstageConditionReason, string) {
+	// Sort pods by creation time, newest first
```
rm3l, line 107: Couldn't this be an issue during rolling updates? For example, after testing a CR reconciled correctly (all 2 desired replicas Ready), then I tried to make it fail (just by enabling the lightspeed flavour), the new pod created fails to start, but when checking the CR status, I can see this :
```yaml
status:
  conditions:
  - lastTransitionTime: "2026-08-23T20:32:11Z"
    message: 2/2 replicas ready
    reason: Deployed
    status: "True"
    type: Deployed
  - lastTransitionTime: "2026-08-23T20:28:02Z"
    message: 'init container "install-dynamic-plugins": CrashLoopBackOff'
    reason: ContainerFailed
    status: "False"
    type: Runtime
```

So I feel like the result is contradictory and might be confusing: fully deployed with 2/2 replicas ready but container failed.

api/current-types.go
```diff
@@ -49,11 +49,23 @@ type (
...
+	// Condition types
+	BackstageConditionTypeDeployed BackstageConditionType = bsv1.BackstageConditionTypeDeployed
+	BackstageConditionTypeRuntime  BackstageConditionType = bsv1.BackstageConditionTypeRuntime
+	BackstageConditionTypeConfig   BackstageConditionType = bsv1.BackstageConditionTypeConfig
+
+	// Deployed condition reasons
 	BackstageConditionReasonDeployed   BackstageConditionReason = bsv1.BackstageConditionReasonDeployed
 	BackstageConditionReasonFailed     BackstageConditionReason = bsv1.BackstageConditionReasonFailed
 	BackstageConditionReasonInProgress BackstageConditionReason = bsv1.BackstageConditionReasonInProgress
-	BackstageConditionReasonIdled      BackstageConditionReason = bsv1.BackstageConditionReasonIdled
```
rm3l, line 56: If this Reason is being dropped, let's also consider updating https://github.com/gazarenkov/rhdh-operator/blob/plugin-status/docs/admin.md#idling-an-instance accordingly, especially how CI and monitoring scripts should distinguish an intentionally idled instance from a deployment failure.

gazarenkov: fixed

internal/controller/backstage_controller.go
gazarenkov: fixed

internal/controller/backstage_status.go
gazarenkov: Good catch

Added staled rollout checking and additional reason to reflect it, now you should see something like:

conditions:                                                                                                                                  
- type: Deployed                                                                                                                           
  status: "False"                                                                                                                          
  reason: RolloutStalled                                                                                                                    
  message: "2/2 replicas ready, 0/2 updated (rollout stalled)" 
  ...
  in this case

rm3l (approved):

Fortune-Ndlovu (approved):
lgtm

PR #3359, opened 2026-08-13
feat: build operator images hermetically in upstream repo [RHIDP-15044]

Fortune-Ndlovu:
Consolidate upstream and downstream Dockerfiles into a single unified Dockerfile that works with both Hermeto (upstream) and Cachi2 (Konflux) hermetic builds. This catches build issues at PR time instead of waiting for downstream Konflux builds.

Fixes: https://redhat.atlassian.net/browse/RHIDP-15044

## PR acceptance criteria

- [x] Tests
- [x] Documentation

```bash
  # Run the full hermetic build (fetches deps + builds image with --network none)
  ./scripts/local-hermeto-build.sh -d . -i localhost/operator:hermetic-test

  # Verify the image exists and the binary runs
  podman images localhost/operator:hermetic-test
  podman run --rm localhost/operator:hermetic-test --help
```


Fortune-Ndlovu:
@qodo regarding Hermeto image tag unpinned this is intentional as of why I skipped it because the team previously decided against pinning `:latest` for tooling images (consistent with rhdh repo pattern)

.github/actions/docker-build/action.yaml
```diff
@@ -0,0 +1,188 @@
...
+        if ! command -v podman &>/dev/null; then
+          echo "podman not found, installing..."
+          sudo apt-get -y update && sudo apt-get -y install podman
+        fi
+        podman --version
+
+    - name: Set up hermetic build variables
+      shell: bash
+      run: |
+        echo "HERMETO_IMAGE=quay.io/konflux-ci/hermeto:latest" >> "$GITHUB_ENV"
```
nickboldt: can we pin to 0.60.1 (from https://quay.io/repository/konflux-ci/hermeto?tab=tags&tag=0.60.1 ) and use renovate to bump this as new releases happen? otherwise you get unexpected updates to hermeto you didn't expect... or you test locally against a cached :latest which is not the true :latest in the wild.

Fortune-Ndlovu: Re: pinning hermeto image https://github.com/redhat-developer/rhdh-operator/pull/3359#pullrequestreview-4928669472, good call. Pinned to `0.60.1` in both the CI composite action and the local build script (897bb5ff). Also added a `# renovate: datasource=docker depName=quay.io/konflux-ci/hermeto` comment hint so Renovate will track and bump it automatically.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-# THIS IS USED BY Konflux builds >= 1.4 with Cachi2 enabled
+# Unified Dockerfile for both upstream (Hermeto) and downstream (Konflux/Cachi2) hermetic builds
```
nickboldt: can we rename to Containerfile here and also downstream?

.github/actions/docker-build/action.yaml
```diff
@@ -0,0 +1,189 @@
...
+        # renovate: datasource=docker depName=quay.io/konflux-ci/hermeto
+        echo "HERMETO_IMAGE=quay.io/konflux-ci/hermeto:0.60.1" >> "$GITHUB_ENV"
+        echo "LOCAL_CACHE_DIR=./hermeto-cache/operator" >> "$GITHUB_ENV"
+
+    - name: Restore hermeto dependency cache
+      id: cache-deps
+      uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      with:
+        path: ./hermeto-cache/operator
+        key: hermeto-deps-${{ inputs.platform }}-${{ hashFiles('go.sum', 'rpms.lock.yaml') }}
```
polasudo: The cache key hashes only go.sum and rpms.lock.yaml
A go.mod-only change (new module path, replace directive, Go version bump) can still hit cache and skip fetch-deps, leaving stale /cachi2/output for a --network=none build.

I would say we can add this one here 
`key: hermeto-deps-${{ inputs.platform }}-${{ hashFiles('go.mod', 'go.sum', 'rpms.lock.yaml') }}
`
WDYT?

Dockerfile
Fortune-Ndlovu: yeah I have tracked this follow-up PR here https://redhat.atlassian.net/browse/RHDHBUGS-3654

.github/actions/docker-build/action.yaml
Fortune-Ndlovu: Adding go.mod to the cache key is a valid point because without a go.sum change would hit stale cache. Thanks

polasudo (approved):
as comments were addressed, LGTM

Dockerfile
gazarenkov: I suppose it also still support non hermetic ('make image-build') way.
Make sense to mantion it here as well?

gazarenkov (approved):

PR #3373, opened 2026-08-20
chore: adopt UBI micro distroless-style image for RHDH operator [RHIDP-15087]

Fortune-Ndlovu:
Replace UBI minimal runtime base with UBI micro using the --installroot pattern. An intermediate rpm-builder stage installs openssl (for FIPS support) into an isolated rootfs, which is overlaid onto a UBI micro base image. This reduces the final image size by ~31% and eliminates the package manager, shell utilities, and other attack surface from the shipped image.

- Fixes: https://redhat.atlassian.net/browse/RHIDP-15087

#### 1. Build the image

```bash
# Using podman
CONTAINER_TOOL=podman make image-build

# Or using docker
make image-build
```

#### 2. Verify image size reduction

```bash
# New image (UBI micro) — expect ~159 MB
podman images quay.io/rhdh/rhdh-rhel9-operator:2.0 --format "{{.Size}}"

# Compare with current main image (UBI minimal) — ~231 MB
# Build from main branch for a direct comparison:
git stash && git checkout main && CONTAINER_TOOL=podman make image-build
podman images quay.io/rhdh/rhdh-rhel9-operator:2.0 --format "{{.Size}}"
git checkout - && git stash pop
```

| Image | Base | Size | Reduction |
|---|---|---|---|
| main (before) | UBI minimal | ~231 MB | — |
| This PR (after) | UBI micro | ~159 MB | **~31%** |

#### 3. Verify distroless properties (no package manager, no attack tools)

```bash
IMG=quay.io/rhdh/rhdh-rhel9-operator:2.0

# No package manager
podman run --rm --entrypoint=/bin/sh $IMG -c 'which dnf 2>/dev/null || echo "dnf: not found"'
podman run --rm --entrypoint=/bin/sh $IMG -c 'which microdnf 2>/dev/null || echo "microdnf: not found"'

# No network attack tools
podman run --rm --entrypoint=/bin/sh $IMG -c 'which curl 2>/dev/null || echo "curl: not found"'
podman run --rm --entrypoint=/bin/sh $IMG -c 'which wget 2>/dev/null || echo "wget: not found"'

# Empty RPM database (no tracked packages)
podman run --rm --entrypoint=/bin/sh $IMG -c 'rpm -qa 2>/dev/null | wc -l'
```

Expected output:
```
dnf: not found
microdnf: not found
curl: not found
wget: not found
0
```

#### 4. Verify OpenSSL is present (FIPS support)

```bash
podman run --rm --entrypoint="" $IMG /usr/bin/openssl version
```

Expected: `OpenSSL 3.5.5 ...`

#### 5. Verify image digests are multi-arch manifest lists

```bash
make validate-image-digests
```

Expected: all digests pass as manifest lists.

#### 6. Verify hermetic build (simulates Konflux pipeline)

```bash
CONTAINER_TOOL=podman make hermetic-build
```

This tests that the `sed`-based Dockerfile transforms in `scripts/local-hermeto-build.sh` correctly match the new `RUN dnf install --installroot` pattern.

#### 7. Deploy and verify operator starts

```bash
# Deploy to a test cluster
make deploy

# Verify the operator pod reaches Running state
kubectl get pods -n backstage-system

# Verify the operator can reconcile a Backstage CR
kubectl apply -f examples/backstage.yaml
kubectl get backstage -A
```

#### 8. Debug in production (no shell in the image)

Since UBI micro has no shell by default, use ephemeral debug containers:

```bash
kubectl debug -it <pod-name> --image=registry.access.redhat.com/ubi9-minimal --target=<container-name>
```

### DIFF
--- a/.github/workflows/update-rpm-lockfile.yaml
+++ b/.github/workflows/update-rpm-lockfile.yaml
@@ -60,8 +60,8 @@ jobs:
 
       - name: Run rpm-lockfile-prototype
         run: |
-          echo "Running '${HOME}/.local/bin/rpm-lockfile-prototype -f ${{ env.DOCKERFILE_PATH }} rpms.in.yaml' in $(pwd)"
-          ${HOME}/.local/bin/rpm-lockfile-prototype -f ${{ env.DOCKERFILE_PATH }} rpms.in.yaml
+          echo "Running '${HOME}/.local/bin/rpm-lockfile-prototype --bare rpms.in.yaml' in $(pwd)"
+          ${HOME}/.local/bin/rpm-lockfile-prototype --bare rpms.in.yaml
 
       - name: Check for lockfile changes
         id: check-lockfile-changes
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,7 @@ COPY --from=rpm-builder /mnt/rootfs /
 COPY LICENSE /licenses/
 
 # Copy manager binary
-COPY --from=builder /opt/app-root/src/manager .
+COPY --from=builder /opt/app-root/src/manager /manager
 
 USER 1001
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,30 +31,33 @@ COPY $EXTERNAL_SOURCE $CONTAINER_SOURCE
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
 
-# Install openssl for FIPS support
-#@follow_tag(registry.redhat.io/ubi9/ubi-minimal:latest)
-# https://registry.access.redhat.com/ubi9-minimal
-FROM registry.access.redhat.com/ubi9-minimal:9.8-1786987521@sha256:8eb2830d0936237fc13a1f2f7e45aecf90d69043380ad167fad0343632937f41 AS runtime
+# Install openssl for FIPS support into an isolated rootfs
+#@follow_tag(registry.redhat.io/ubi9/ubi:latest)
+# https://registry.access.redhat.com/ubi9/ubi
+FROM registry.access.redhat.com/ubi9/ubi:9.8-1786985871@sha256:5426a8f45e80a07168a30ea24d84f266094b3756624a5508cc53927e6ee39e09 AS rpm-builder
+RUN mkdir -p /mnt/rootfs
+RUN dnf install --installroot /mnt/rootfs \
+    openssl \
+    --releasever 9 --setopt=install_weak_deps=0 --nodocs -y && \
+    dnf --installroot /mnt/rootfs clean all && \
+    rm -rf /mnt/rootfs/var/cache/* /mnt/rootfs/var/log/* /mnt/rootfs/tmp/*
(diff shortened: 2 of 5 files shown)


Fortune-Ndlovu:
@qodo 

Finding #1: UBI stage version skew this is Not actionable

The different build IDs between `rpm-builder` (`ubi9/ubi:9.8-1786985871`) and the runtime stage (`ubi9/ubi-micro:9.8-1786321990`) are **expected and correct**. These are independent images rebuilt by Red Hat on different schedules, they never share build IDs.

The latest available tags confirm we are already pinned to the newest version of each:

| Image | Latest 9.8 tag | Catalog |
|-------|----------------|---------|
| `ubi9/ubi` | `9.8-1786985871` (our pin) | [Red Hat Ecosystem Catalog](https://catalog.redhat.com/en/software/containers/ubi9/ubi/615bcf606feffc5384e8452e) |
| `ubi9/ubi-micro` | `9.8-1786321990` (our pin) | [Red Hat Ecosystem Catalog](https://catalog.redhat.com/en/software/containers/ubi9/ubi-micro/615bd480b75f4efe6f60e1a0) |

There is no `ubi-micro:9.8-1786985871` tag it does not exist. The `--installroot` COPY overlay pattern means the installroot's packages fully replace any overlapping files in the ubi-micro base, so there is no version conflict at runtime. This is the same pattern used in [`plugin-installer/Dockerfile.skopeo`](https://github.com/redhat-developer/rhdh-operator/blob/main/plugin-installer/Dockerfile.skopeo). Renovate will independently bump each image when Red Hat publishes new builds.

Finding #3: Installroot releasever mismatch Not valid

`--releasever 9` is correct. UBI repositories are organized by **major version** (`9`), not minor version (`9.8`). The repo URLs in [`rpms.in.yaml`](https://github.com/redhat-developer/rhdh-operator/blob/main/rpms.in.yaml) confirm this structure:

```
https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/$basearch/baseos/os/
```

Using `--releasever 9.8` would fail because no repositories exist at that path. The reference implementation in [`plugin-installer/Dockerfile.skopeo`](https://github.com/redhat-developer/rhdh-operator/blob/main/plugin-installer/Dockerfile.skopeo#L19) also uses `--releasever 9`.

No code changes needed for either finding.

Fortune-Ndlovu:
## Why the lockfile and workflow changed

The old runtime image (ubi9-minimal) came with ~150 packages already installed. When we ran `rpm-lockfile-prototype -f Dockerfile`, it saw those packages and only locked the 4 extra ones needed for openssl.

Our new approach uses `dnf install --installroot /mnt/rootfs`, which installs into a completely empty filesystem — nothing is pre-installed. So openssl now needs all of its dependencies (glibc, bash, ca-certificates, coreutils, etc.) explicitly listed in the lockfile. That's why it grew from 4 entries to 142.

The `--bare` flag tells `rpm-lockfile-prototype` to resolve against an empty rootfs, which is exactly what `--installroot` does at build time. Without this change, the hermetic build failed because the lockfile was missing packages that `--installroot` needed to install from scratch.

The workflow change in `update-rpm-lockfile.yaml` is the same update — the weekly automated lockfile refresh now uses `--bare` so it stays consistent.

Fortune-Ndlovu:
Reopening from fork branch. See replacement PR.

PR #3473, opened 2026-09-03
fix: restrict ref:// and {{inherit}} resolution to OCI base plugins

jonkoops:
Both `ref://` and `{{inherit}}` now validate that the matched base plugin has an `oci://` package URL. Attempting to resolve against a base plugin with `https://`, `http://`, or `./` returns an error. Direct links for all formats continue to work.

Also rewrites the "Plugin URL References" documentation to explain how name matching works (annotated OCI URL diagram, concrete before→after examples) and corrects the init-container comparison: `ref://` matching is now consistent across both implementations, while `{{inherit}}` still differs (operator uses name-based matching, init container uses full-URL key matching).

[RHIDP-15875](https://redhat.atlassian.net/browse/RHIDP-15875)

### DIFF
--- a/docs/dynamic-plugins.md
+++ b/docs/dynamic-plugins.md
@@ -97,38 +97,66 @@ More details in [Catalog Entities Extraction](https://github.com/redhat-develope
 
 ## Plugin URL References
 
-The operator optionally supports special URL reference syntax in plugin package URLs, allowing users to reference plugins from the default configuration by name.
+The operator supports special URL reference syntax in plugin package URLs, allowing users to reference plugins from the default configuration by name instead of specifying a full OCI image reference. Both reference types only resolve to base plugins with `oci://` package URLs.
 
 TODO: document Operator Dynamic Plugins processing mode
 
 **Operator behavior:**
 - The operator resolves all references during ConfigMap merge (before passing to the init container)
 - If a reference cannot be resolved, the operator returns an error and the Backstage CR will not reconcile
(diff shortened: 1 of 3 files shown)


docs/dynamic-plugins.md
```diff
@@ -97,38 +97,74 @@ More details in [Catalog Entities Extraction](https://github.com/redhat-develope
...
+    # → resolves to: oci://quay.io/rhdh/backstage-plugin-catalog@sha256:abc123!backstage-plugin-catalog
     pluginConfig:
       # your config overrides
 ```
 
-### Inherit Reference (`:{{inherit}}`)
+### Inherit Reference (`{{inherit}}`)
 
-Look up a plugin by name and use its full package URL from the default configuration. The registry/path in your URL is ignored - only the plugin name matters for matching.
+Write an `oci://` URL with `:{{inherit}}` in place of the tag or digest. The operator extracts the plugin name from your URL, looks it up in the default plugin list, and substitutes the full OCI URL from the match. The registry and path you write are ignored; only the plugin name (last path component) is used for lookup.
```
gazarenkov, line 139: !custom-path ignored as well

gazarenkov (approved):

docs/dynamic-plugins.md
jonkoops: Removed the `!plugin-path` override section and all `!plugin-path` suffixes from the examples.

rm3l (approved):

PR #3507, opened 2026-09-10
 Add inline dynamic plugins configuration to Backstage CR

gazarenkov:
## Description
 This PR adds support for configuring dynamic plugins directly in the Backstage CR using the spec.application.dynamicPlugins field, eliminating the need for a separate ConfigMap in simple use cases.

Users can now configure plugins directly in the Backstage CR:

  ```
  kind: Backstage
  metadata:
    name: my-backstage
  spec:
    application:
      dynamicPlugins:
        - package: 'ref://backstage-plugin-kubernetes'
          pluginConfig:
            kubernetes:
              serviceLocatorMethod:
                type: multiTenant
        - package: 'ref://backstage-plugin-github-actions'
          enabled: true
        - package: 'oci://quay.io/my-org/custom-plugin@sha256:abc123'
```

https://redhat.atlassian.net/browse/RHIDP-15060

## PR acceptance criteria

- [x] Tests
- [x] Documentation

For quick start use:
examples/inline-dynamic-plugins.yaml
example

### CODE REVIEW

File: pkg/model/dynamic-plugins.go [hunk 1]
```diff
@@ -353,10 +403,13 @@ func MergePluginsData(firstData, secondData string) (string, error) {
...
 			if plugin.Enabled != nil {
 				existingPlugin.Enabled = plugin.Enabled
-				existingPlugin.Disabled = false
 			} else if plugin.Disabled {
 				existingPlugin.Disabled = true
 				existingPlugin.Enabled = nil
+			} else {
+				// User added this plugin to overlay without specifying enabled/disabled
+				// Default to enabled
+				existingPlugin.Enabled = ptr.To(true)
```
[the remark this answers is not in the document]
gazarenkov: Enabled by default is what is documented already, so not an issue

api/v1alpha5/backstage_types.go
```diff
@@ -127,9 +129,17 @@ type Application struct {
...
+	// Mutually exclusive with DynamicPlugins field.
 	// +optional
 	DynamicPluginsConfigMapName string `json:"dynamicPluginsConfigMapName,omitempty"`
 
+	// Inline dynamic plugins configuration.
+	// Allows configuring dynamic plugins directly in the Backstage CR without requiring a separate ConfigMap.
+	// Each plugin can specify its package reference, enabled state, plugin-specific configuration, and integrity checksum.
+	// Mutually exclusive with DynamicPluginsConfigMapName field.
+	// +optional
+	DynamicPlugins []DynamicPluginConfig `json:"dynamicPlugins,omitempty"`
```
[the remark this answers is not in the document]
gazarenkov: Expected, v1alpha4 will ignore new field

pkg/model/dynamic-plugins.go
```diff
@@ -107,29 +109,39 @@ func (p *DynamicPlugins) addToModel(model *BackstageModel, backstage api.Backsta
...
+	// Determine source of user-defined plugins (CR inline or external ConfigMap)
+	// CEL validation ensures these are mutually exclusive
+	var userPluginsData string
 
-		// if the ConfigMap is set but does not have the data or expected key
+	if backstage.Spec.Application != nil && len(backstage.Spec.Application.DynamicPlugins) > 0 {
+		// Get plugins from CR inline configuration
+		data, err := convertInlinePlugins(backstage.Spec.Application.DynamicPlugins)
+		if err != nil {
+			return fmt.Errorf("failed to convert inline plugins: %w", err)
```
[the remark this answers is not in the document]
gazarenkov: The configMap is not expected to be changed manually, changing CR will cause refreshing

crdify.yaml
```diff
@@ -0,0 +1,8 @@
+# crdify configuration
+# Allow description changes and unhandled XValidations for alpha APIs (v1alpha4, v1alpha5)
+
+validations:
+  - name: description
+    enforcement: None
+
+unhandledEnforcement: None
```
rm3l:
```suggestion
unhandledEnforcement: Warn
```

I would suggest reporting those as warnings (at minimum). Otherwise, they are completely hidden.
See my other comment below about the CRD upgrade concerns: https://github.com/redhat-developer/rhdh-operator/pull/3507#discussion_r4005311184

api/v1alpha5/backstage_types.go
```diff
@@ -112,6 +112,8 @@ type Monitoring struct {
 	Enabled bool `json:"enabled,omitempty"`
 }
 
+// Application defines configuration for the Backstage application.
+// +kubebuilder:validation:XValidation:rule="!(has(self.dynamicPlugins) && size(self.dynamicPlugins) > 0 && has(self.dynamicPluginsConfigMapName) && size(self.dynamicPluginsConfigMapName) > 0)",message="dynamicPlugins and dynamicPluginsConfigMapName are mutually exclusive"
```
rm3l: I suspect that, even with the `crdify.yaml` file that we have, the OCP OLM v1 preflight checks will block this upgrade as they also use crdify directly (one of the reasons why we introduced it too in our CI to catch these issues earlier); and it looks like the default config in the OCP OLM v1 controller would handle this addition as an error. See https://github.com/openshift/operator-framework-operator-controller/blob/433f3cc944ee4f7995a5d0b444aecb77c3891176/internal/operator-controller/rukpak/preflights/crdupgradesafety/crdupgradesafety.go#L117

I am not sure there is a way for operator teams to provide their own crdify config in the bundle; the only user control is a setting toggle to disable the entire preflight check on their ClusterExtension: https://docs.okd.io/4.21/extensions/ce/crd-upgrade-safety.html#disabling-crd-preflight_crd-upgrade-safety

So the XValidation addition here on `spec.application` would be flagged as an unhandled change and the upgrade would fail. And even if https://github.com/kubernetes-sigs/crdify/issues/50 is addressed to "handle" it in crdify, it would still see XValidation additions as "breaking changes", as depicted in that issue.

As an option to avoid forcing customers to disable preflight safety checks, maybe we should drop this CEL XValidation marker comment and enforce mutual exclusivity in the controller's Go code instead? Sure, the user experience would be different: CR accepted then error in status vs CR rejected at apply time, but that would avoid the OLM upgrade blocker entirely, I believe.
Otherwise, this should be documented as potential breaking change for 2.1.0: administrators should temporarily disable preflight checks before upgrading the RHDH operator (which doesn't look nice either).
WDYT?

api/v1alpha5/backstage_types.go
rm3l, line 140:
```suggestion
	DynamicPlugins *[]DynamicPluginConfig `json:"dynamicPlugins,omitempty"`
```

Would it make sense to use a pointer to distinguish "unset" vs "empty" cases? Better to think about this now that the field is new, otherwise it might become harder to change later as it might fail the CRD upgrade check..

pkg/model/dynamic-plugins_test.go
```diff
@@ -586,7 +589,7 @@ plugins:
 
 		plugin := findPluginByPackage(config.Plugins, "./plugin-a")
 		assert.NotNil(t, plugin)
-		assert.True(t, plugin.IsDisabled(), "base disabled: true should be preserved when overlay omits both fields")
+		assert.True(t, !plugin.IsDisabled(), "enabled by default")
```
rm3l:
```suggestion
		assert.False(t, plugin.IsDisabled(), "enabled by default")
```

nit: double negation

gazarenkov: fixed

api/v1alpha5/backstage_types.go
gazarenkov: Do you have any concrete case in mind? 
Currently it behaves exactly the same with empty and nil spec

api/v1alpha5/backstage_types.go
gazarenkov: Your point is valid, even if it is NOT a breaking change, crdfy tool is too dumb to recognize it so far...
So moved validation to controller

crdify.yaml
gazarenkov: fixed

api/v1alpha5/backstage_types.go
rm3l: Yeah, I agree this specific rule should not be flagged as breaking change, but understand it might not be that straightforward for crdify to detect this from the XValidation.

api/v1alpha5/backstage_types.go
rm3l: I was thinking about users who might want to start with zero plugin, not even the default ones from the default catalog. But that might not be an issue.

I did this by specifying an empty list of `includes` in the dynamic plugins ConfigMap then adding the specific set of plugins I wanted to test, like so:
```yaml
includes: []
plugins:
  - package: ...
```

This makes me realize that the list of `includes` is not exposed here. Is that intentional?

gazarenkov: I see, valid point, but I am not sure if it helps the same way as for "includes". We may do that so spec.app.dynaPlugins[] will force ignoring defaults but then how to add needed plugins?

Actually, we can exclude defaults by:
- on test removing (or clean) profile/rhdh/local-test/dynamic-plugins.yaml (just modified the script to not fail)
- on prod deleting created by default devhubplugincatalog CR

WDYT?

rm3l: Sounds good, thanks.

rm3l (approved):

PR #3508, opened 2026-09-11
fix(e2e): remove route deletion check and skip ephemeral volume test [RHIDP-14647]

Fortune-Ndlovu:
## Description

This PR fixes e2e failures when running against OpenShift with OLM v1 by aligning route assertions with the operator’s behavior (disabling a route stops managing it but does not delete it immediately; cleanup happens via owner-reference GC when the Backstage CR is deleted) and by removing the incorrect `raw-runtime-config` skip. Re-testing on OCP 4.22 showed generic ephemeral volumes work, but PVC provisioning can be slow on `WaitForFirstConsumer` storage classes, so the README documents an optional longer reachability timeout.

- Fixes https://redhat.atlassian.net/browse/RHIDP-14647

## PR acceptance criteria

- [x] Tests
- [x] Documentation

```bash
make test-e2e \
  BACKSTAGE_OPERATOR_TEST_MODE=rhdh-latest \
  BACKSTAGE_OPERATOR_TESTS_APP_REACHABILITY_TIMEOUT=15m
```

```bash
make test-e2e \
  BACKSTAGE_OPERATOR_TEST_MODE=rhdh-latest \
  GINKGO_TEST_ARGS='--focus "raw-runtime-config"' \
  BACKSTAGE_OPERATOR_TESTS_APP_REACHABILITY_TIMEOUT=15m
```


tests/e2e/e2e_test.go
```diff
@@ -116,6 +116,11 @@ var _ = Describe("Backstage Operator E2E", func() {
 					if tt.isForOpenshift && !helper.IsOpenShift() {
 						Skip("Skipping OpenShift-only test on non OCP platform")
 					}
+					if tt.name == "raw-runtime-config" {
+						if os.Getenv("USE_EXISTING_CLUSTER") == "true" || os.Getenv("SKIP_RAW_RUNTIME_CONFIG_TEST") == "true" {
```
rm3l, line 120: Skipping for `USE_EXISTING_CLUSTER` seems too broad IMO, as it would silently drop coverage on any type of cluster. Can't this be gated only on `SKIP_RAW_RUNTIME_CONFIG_TEST`?

tests/e2e/e2e_test.go
```diff
@@ -116,6 +116,11 @@ var _ = Describe("Backstage Operator E2E", func() {
...
+							Skip("Skipping raw-runtime-config: requires ephemeral volume controller support")
```
rm3l, line 121: Are you sure the root cause is the lack of ephemeral volume controller support? Generic ephemeral volumes have been GA [since Kubernetes 1.23](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes) and is also documented on OCP 4.22: https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/storage/generic-ephemeral-volumes-1

Wasn't the issue more related to a missing/misconfigured storage provisioning on your OCP cluster?

tests/e2e/e2e_test.go
Fortune-Ndlovu: Agreed, removed the `USE_EXISTING_CLUSTER` gate entirely. The `raw-runtime-config` test now always runs with the rest of the suite.

I also tightened the route assertions: we no longer expect immediate route deletion when `route.enabled=false`, but we still verify the route is removed after CR deletion whenever one was created during the test (including the enable-then-disable case).

tests/e2e/e2e_test.go
Fortune-Ndlovu: Good catch, I re-tested on OCP 4.22 with OLM v1 and the root cause is not missing ephemeral volume controller support.

`raw-runtime-config` passes without any skip. Pods can sit in `Pending` for a while with `waiting for ephemeral volume controller to create the persistentvolumeclaim ... not found` while the generic ephemeral volume PVC is being provisioned (our cluster uses `gp3-csi` with `WaitForFirstConsumer`). With the default `5m` reachability timeout the test was flaky; with `BACKSTAGE_OPERATOR_TESTS_APP_REACHABILITY_TIMEOUT=15m` the full e2e suite passed (7/7) in ~22 minutes.

I've dropped the skip and documented the optional timeout in `tests/e2e/README.md` instead of attributing this to missing controller support.

rm3l (approved):

rm3l (approved):

PR #3557, opened 2026-09-17
fix: replace `/build-images` slash command with secure build/push split

rm3l:
## Description

Replace the `issue_comment`-triggered `/build-images` workflow with a
`pull_request`-triggered build + `workflow_run`-triggered push, following
the pattern already used in the must-gather repo.

The build workflow runs fork code with no secrets (`contents: read` only)
and saves operator and bundle images as artifacts. The push workflow
downloads artifacts, validates metadata, pushes images, and builds the
catalog inline using `opm` (since `opm render` needs the bundle pushed
to the registry first) — no fork Makefile ever runs with registry
credentials.

- Fixes PSIRTSUPT-22717

## PR acceptance criteria

- [x] Tests
- [x] Documentation

- Open a PR with a Makefile/Go change — build triggers, produces `pr-images` artifact
- On build success — push workflow fires, pushes 3 images to Quay, comments on PR
- Open a docs-only PR — build skips (no changed files), push skips (no artifact)
- Build failure — `notify-failure` job posts failure comment on PR

## Container Images

Container images are built and pushed to Quay automatically when relevant files change.
Image links will be posted in a PR comment once the push completes.


.github/workflows/pr-container-push.yaml
```diff
@@ -0,0 +1,422 @@
...
+            const { data: { artifacts } } = await github.rest.actions.listWorkflowRunArtifacts({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id: runId,
+            });
+            const found = artifacts.some(a => a.name === 'pr-images');
+            console.log(`Artifact 'pr-images' for run ${runId}: ${found ? 'found' : 'not found'}`);
+            core.setOutput('exists', found);
+
+      - name: Download artifact from build workflow
```
nickboldt, line 112: do we need to use artifacts? can we just push the PR build directly to quay under https://quay.io/repository/rhdh-community/rhdh-pr?tab=tags or https://quay.io/repository/rhdh-community/rhdh-operator-pr?tab=tags so it's separate from "real" repos where we push CI and GA builds? See simplified workflow proposal for rhdh repo in https://github.com/redhat-developer/rhdh/pull/5437/changes#top using new PR user/token.

rm3l: I took a quick look at your rhdh PR, but [this](https://github.com/redhat-developer/rhdh/blob/90d93009a21982d20706eb924502fd5c1d52e769/.github/workflows/pr-build-image.yaml#L39-L43) means that it will fail right away for fork PRs, no? This might be an issue for us and contributors.
I've left a comment: https://github.com/redhat-developer/rhdh/pull/5437#pullrequestreview-5240405340

We can push to a new `*-pr` repo and that would minimize the blast radius to only PR images, but still, the risk in PSIRTSUPT-22717 was that we could be running `make` command after checking out untrusted PR code IIUC.

nickboldt (approved):
works for me

nickboldt:
So... I guess we don't need a new token for pushing to a new repo suffixed with `-pr` as in https://github.com/redhat-developer/rhdh/pull/5437 ?

rm3l:
> So... I guess we don't need a new token for pushing to a new repo suffixed with `-pr` as in [redhat-developer/rhdh#5437](https://github.com/redhat-developer/rhdh/pull/5437) ?

In the remediation guidance on both https://redhat.atlassian.net/servicedesk/customer/portal/68/PSIRTSUPT-22717 and https://redhat.atlassian.net/servicedesk/customer/portal/68/PSIRTSUPT-22719, it is mentioned that we should assume compromise and rotate the tokens, then review access logs, and audit all tags and manifests for unauthorized changes. So I guess it would be safer to rotate the current quay token anyway as a starting point? WDYT?

PR #3569, opened 2026-09-18
feat: upgrade default PostgreSQL to 18

zdrapela:
## Summary
- update default PostgreSQL images from 15 to 18
- add PostgreSQL dump, volume replacement, and restore coverage to Operator upgrade E2E tests
- run the migration flow in nightly main upgrades and a dedicated pull-request check

This upgrade path was tested with the full E2E test suite in https://github.com/redhat-developer/rhdh/pull/5145

## Verification
- `make lint`
- `make test OPERATOR_DP_PROCESSING=false`
- focused restore validation tests
- `actionlint` for the new pull-request workflow

Jira: https://redhat.atlassian.net/browse/RHIDP-13959


tests/e2e/operator_cleanup_test.go
rm3l, line 1: Do we really need this unit test testing a helper defined in the E2E suite? `captureOperatorLogsBeforeCleanup` is a 3 lines function..

tests/e2e/postgresql_upgrade_restore_test.go
rm3l, line 1: Same for this btw. This unit test file doesn't seem necessary IMO.

.github/workflows/pr-postgresql-upgrade-test.yaml
rm3l, line 1: Do we really need this on every PR? Maybe just the nightly upgrade workflow above is enough? We could maybe just limit it for now to the `release-1.10 => main` upgrade path today (and later include `1.10 => release-2.1` too)?

.github/workflows/nightly-postgresql-upgrade-test.yaml
zdrapela: Agreed, it's quite a lengthy test, and nightly is good enough. 
The `1.10 => release-2.1` should be the ultimate target, and maybe we can later get rid of `release-1.10 => main` in the future (when most of the customers that may be using the internal DB migrate).

tests/e2e/operator_cleanup_test.go
zdrapela: You're right. 
I'm still fighting with the cheap models in combination with skills, where they insist on writing unit tests for everything 😅

rm3l (approved):

zdrapela:
Trying to comment to unblock `tide`

